PR #1, opened 2017-10-30
(MODULES-5844) Rename dsc module namespace to dsc_lite

jpogran:
This PR renames the PowerShellManager namespace to dsc_lite. This is done
by moving all the class files either in a root folder or a folder named
dsc and moving them to a dsc_lite folder. In cases where it was a
template file, an attempt was made to move the template closer to the
ruby files using it, instead of a seperate template dir structure.

This PR also removes all files related to the previous dsc module, namely all built types, DSC Resources, build files, and other associated files. These commits could be moved to another PR, but were necessary to keep the namespace work sane, so it was kept in this PR but split among separate commits.

Note:

* I stubbed out the new `dsc` type in order to keep spec tests working, it still needs to be implemented
* I expect all acceptance tests to fail, as they expect the full dsc types present to work. A new ticket will be raised to decide whether to edit all tests inline to use the new type to keep the tests relevant, or to remove all but one of two that can prove the new type works
* `lib/puppet/provider/templates/dsc/invoke_dsc_resource.ps1.erb` is not touched in this PR as there is significant work to be done to it to make it work with the new dsc type


Iristyle:
I wonder if we should put up a PR to trash all of the content up-front.

jpogran:
@Iristyle It would make things easier for me in renaming. I was pseudo following our order in jira. In testing locally, with the existing dsc module present, I'm having to rename everything anyway

Iristyle:
Handled in #3, #4 and #8 so closing

PR #2, opened 2017-11-08
(WIP) (MODULES-5842) Generic DSC invoker

jpogran:
Note: PR #1 should be merged first

This adds a generic `dsc` type with the following signature:

Caveat: Names of all parameters/properties are not final

```
dsc{'WindowsFeature':
  dsc_name      => 'WindowsFeature',
  module_name   => 'PSDesiredStateConfiguration',
  dsc_properties => {
    'ensure' => 'present',
    'name'   => 'Web-Server',
  }
}
```

I know ModuleName was supposed to be added in MODULES-5841, but you cannot use `Invoke-DscResource` without the name of the module.  This can be overloaded with the path to the module without having to change the puppet type.

Both `title` and `dsc_name` are the same right now, with title being required and `dsc_name` not being used. Still deciding whether to use `dsc_name` or not.

The `dsc_properties` parameter has the hash of properties to pass to the `Invoke-DscResource` cmdlet. This is a naive implementation right now, and does not handle PSCredentials or any typed property. Handling this is a TODO at the moment.


lib/puppet/type/dsc.rb
```diff
@@ -4,6 +4,50 @@
...
+    desc "wakka_module_name"
+    isrequired
+    validate do |value|
+      unless value.kind_of?(String)
+        fail("Invalid value '#{value}'. Should be a string")
+      end
+    end
+  end
+
+  newparam(:dsc_name) do
```
Iristyle: @jpogran Regarding the question of whether or not `dsc_name` is necessary that you asked in the PR message... it is.  Given `title` is intended to uniquely identify an individual resource, `title` can't be the name of the DSC type, else you would be limited to a single instance of that resource in the catalog.

lib/puppet/provider/templates/dsc_lite/invoke_dsc_resource.ps1.erb
```diff
@@ -28,46 +23,19 @@ $response = @{
...
-    <%= name %> = <%= value %>
-<% end -%>
-  }<% if resource.respond_to?(:dscmeta_module_version) %>
-  ModuleName = @{
-    ModuleName      = <%= "\"#{vendored_modules_path}/#{resource.dscmeta_module_name}/#{resource.dscmeta_module_name}.psd1\"" %>
-    RequiredVersion = <%= "\"#{resource.dscmeta_module_version}\"" %>
-  }
-<% else %>
-  ModuleName = <%= "\"#{resource.dscmeta_module_name}\"" %>
+Name       = '<%= resource.name %>'
```
Iristyle: I see where the confusion around the acceptance tests come from now.

`dsc_lite` module is intended to be the runtime integration that the `dsc` module consumes, not only the new generic type wrapper.  The generic type wrapper should be additive to what's already been done.

That allows the `dsc` module (and other 3rd party modules) to take a dependency on this module, but just provide content. With all wrapper modules leveraging the same runtime integration, it allows fixes to land in just this `dsc_lite` module rather than being copied into each generated module. Therefore we need to maintain backwards compat ... simplest way is probably to put this code into a new file.

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -184,6 +191,15 @@ def self.format_dsc_value(dsc_value)
     end
   end
 
+  def self.format_dsc_lite(dsc_value)
+    case
+    when dsc_value.class.name == 'Hash'
+      dsc_value.collect{|k, v| format_dsc_value(k) + ' = ' + format_dsc_value(v) + ';' }.join("\n")
```
Iristyle: In our `format_dsc_value` helper, we handle strings by escaping quotes, which AFAIK has worked fine for us so far.

Do you think we need to worry about multi-line strings (i.e. should we use here-strings)? Is that something you've seen with arbitrary DSC resources?

Iristyle:
Work has moved to #12

PR #3, opened 2017-11-09
(MODULES-5968) Generated types / vendored code removal

Iristyle:
Supersedes #1 

**UPDATE: This has been cleaned up / restricted to just MODULES-5968 removals for the sake of separate efforts / merges.  Re-namespacing is in #4 and build task removals are in #5** 

Original differences from that PR (now spanning into the #4 / #5):

* Consolidates removal of generated types, their vendored content + corresponding tests - from the perspective of tests, these should be removed concurrently
* Updated `NOTICE` / `.project` with `dsc_lite`
* Removed additional LCM content - README docs, test helpers, acceptance tests - by keeping f7b15d7 isolated, we can apply it to the DSC repo as well given they have the same history
* Updated `.fixtures.yml` / acceptance test helpers to not install `stdlib` / `powershell` as module deps (were only necessary for LCM manifest)
* Keeps rake tasks / type building helpers ... these will be necessary to keep for the time being - removal is a separate activity
* Updated existing type generation ERB template with correct namespaces - will be necessary to compare old generation to new generation using these for the time being
* Remove `uses_win32console` fact that's not used
* Namespace `invoke_dsc_resource.ps1.erb` / move file to `dsc_lite`
* Keeps `dsc_file`, `dsc_service`, `dsc_xservice` (and vendored `xservice` content from `xPSDesiredStateConfiguration` including `CommonResourceHelpers.psm1`) to keep tests passing (can later swap in our `dsc_puppetfakeresource` or similar for sake of keeping tests valid / passing) - this will need to be removed prior to shipping this module
* Keeps `provider/powershell_spec.rb` running against `dsc_file` - stubbing in new `dsc` type is a separate activity that can be handled in #2
* Fixes requires inside `dsc_puppetfakeresource` so that tests pass
* Removals were done under the wrong ticket MODULES-5839 / filed new ticket MODULES-5968


Iristyle:
Ran each commit through `rake spec`, `integration` and `acceptance` helper scripts. Everything passes along the way, and I think this is good to go.

Will post results when I get to the office.

PR #4, opened 2017-11-15
(MODULES-5844) Move all code to dsc_lite namespace

Iristyle:
Builds on #3, which has now been reduced to removing nearly all vendored content / acceptance tests

This commit moves the PowerShellManager class and dependent files from
the `dsc` namespace to the `dsc_lite` namespace. Where it was possible,
template files were moved closer to the classes that used them. In all
cases a namespace was added to all types and classes to differentiate
the new module from any prior dsc module installs. Changes include:

 * Ruby PuppetX::Dsc -> PuppetX::DscLite
 * Ruby PuppetX::PuppetLabs::Dsc -> PuppetX::PuppetLabs::DscLite
 * renaming Puppet type base_dsc -> base_dcslite
 * using a dsc_lite path to namespace in the filesystem
 * updating existing wrappers (like the fake resource) to use the
   new names and paths


lib/facter/powershell_version.rb
```diff
@@ -1,10 +1,10 @@
 require 'pathname'
-require Pathname.new(__FILE__).dirname + '../' + 'puppet_x/puppetlabs/powershell_version'
+require Pathname.new(__FILE__).dirname + '../' + 'puppet_x/puppetlabs/dsc_lite/powershell_version'
```
Iristyle: Note that `powershell_version` still conflicts on a filename and fact name basis with all the other modules that do this - `powershell`, `dsc` and `iis` (so far). Fortunately we haven't been bit by this given all these modules implement the fact the same way, but something to think about for another day.

PR #12, opened 2017-12-18
(MODULES-5842) Implement Generic DSC Invoker

jpogran:
This commit adds the `dsc` type and provider, which allows executing
DSC Resources without the build system used by the previous pupppet dsc
module.

This is accomplished by adding a new type that accepts a DSC Resource
Name and Module Name, and a un-typed hash of properties. This is passed
as is to `Invoke-DscResource`.

Unit tests for the type are added, manual verification of these changes
are to be done as automated tests are to be added in subsequent tickets

### DIFF
--- a/lib/puppet/provider/base_dsc_lite/powershell.rb
+++ b/lib/puppet/provider/base_dsc_lite/powershell.rb
@@ -3,6 +3,7 @@ require 'json'
 if Puppet::Util::Platform.windows?
   require_relative '../../../puppet_x/puppetlabs/dsc_lite/powershell_manager'
   require_relative '../../../puppet_x/puppetlabs/dsc_lite/compatible_powershell_version'
+  require_relative '../../../puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter'
 end
 
 Puppet::Type.type(:base_dsc_lite).provide(:powershell) do
@@ -57,6 +58,12 @@ EOT
 
   def dsc_parameters
     resource.parameters_with_value.select do |p|
+      p.name.to_s =~ /dsc_/
+    end
+  end
+  
+  def dsc_property_param
+    resource.parameters_with_value.select{ |pr| pr.name == :dsc_resource_properties }.each do |p|
       p.name.to_s =~ /dsc_/
     end
   end
@@ -183,6 +190,10 @@ EOT
       fail "unsupported type #{dsc_value.class} of value '#{dsc_value}'"
     end
   end
+  
+  def self.format_dsc_lite(dsc_value)
+    PuppetX::PuppetLabs::DscLite::PowerShellHashFormatter.format(dsc_value)
+  end
 
   def self.escape_quotes(text)
     text.gsub("'", "''")
@@ -195,7 +206,10 @@ EOT
   def self.ps_script_content(mode, resource, provider)
     dsc_invoke_method = mode
     @param_hash = resource
-    template = ERB.new(File.new(template_path + "/invoke_dsc_resource.ps1.erb").read, nil, '-')
+    template_name = resource.type == :dsc ?
+      '/invoke_generic_dsc_resource.ps1.erb' :
+      '/invoke_dsc_resource.ps1.erb'
+    template = ERB.new(File.new(template_path + template_name).read, nil, '-')
     template.result(binding)
   end
 end
--- a/lib/puppet/type/dsc.rb
+++ b/lib/puppet/type/dsc.rb
@@ -1,0 +1,63 @@
+require 'pathname'
+
+Puppet::Type.newtype(:dsc) do
+  require Pathname.new(__FILE__).dirname + '../../' + 'puppet/type/base_dsc_lite'
+  require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_lite/dsc_type_helpers'
+
+  ensurable do
+    newvalue(:exists?) { provider.exists? }
+    newvalue(:present) { provider.create }
+    newvalue(:absent)  { provider.destroy }
+    defaultto { :present }
+  end
+
+  newparam(:name, :namevar => true) do
+    desc "Name of the declaration"
+    validate do |value|
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
+      end
+      fail("#{value} is not a valid #{self.name.to_s}") unless value =~ /^[a-zA-Z0-9\.\-\_\'\s]+$/
+    end
+  end
+  
+  newparam(:dsc_resource_name) do
+    desc "DSC Resource Name"
+    isrequired
+    validate do |value|
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
+      end
+      fail "#{self.name.to_s} should be a String" unless value.is_a? ::String
+    end
+  end
+
+  newparam(:dsc_resource_module_name) do
+    desc "DSC Resource Module Name"
+    isrequired
+    validate do |value|
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
+      end
+      fail "#{self.name.to_s} should be a String" unless value.is_a? ::String
+    end
+  end
+
+  newparam(:dsc_resource_properties, :array_matching => :all) do
+    desc "DSC Resource Properties"
+    isrequired
+    validate do |value|
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
+      end
+      fail "#{self.name.to_s} should be a Hash" unless value.is_a? ::Hash
+    end
+  end
+end
+
+Puppet::Type.type(:dsc).provide :powershell, :parent => Puppet::Type.type(:base_dsc_lite).provider(:powershell) do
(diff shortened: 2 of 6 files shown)


lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
```diff
@@ -0,0 +1,59 @@
...
+            self.format_array(dsc_value)
+          when dsc_value.class.name == 'Hash'
+            self.format_hash(dsc_value)
+          else
+            fail "unsupported type #{dsc_value.class} of value '#{dsc_value}'"
+          end
+
+        end
+
+        def self.format_string(value)
```
Iristyle: I think these can all be private as there are no other callers and all testing is against `.format`

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+        
+        def self.escape_quotes(text)
+          text.gsub("'", "''")
+        end
+
+      end
+    end
+  end
+end
```
Iristyle: nit - missing newline for some reason

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -183,6 +190,10 @@ def self.format_dsc_value(dsc_value)
       fail "unsupported type #{dsc_value.class} of value '#{dsc_value}'"
     end
   end
+  
+  def self.format_dsc_lite(dsc_value)
```
Iristyle: Is the code here the same as `format_dsc_value` above?

spec/integration/puppet_x/puppetlabs/powershell_hash_formatter_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+require 'spec_helper'
+require 'puppet/type'
+require 'puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter'
+
+describe PuppetX::PuppetLabs::DscLite::PowerShellHashFormatter do
+  before(:each) do
+    @formatter = PuppetX::PuppetLabs::DscLite::PowerShellHashFormatter
+  end
+
+  describe "formating ruby hash to powershell hash string" do
```
Iristyle: nit `formatting`

lib/puppet/type/dsc.rb
```diff
@@ -0,0 +1,63 @@
...
+    isrequired
+    validate do |value|
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
+      end
+      fail "#{self.name.to_s} should be a String" unless value.is_a? ::String
+    end
+  end
+
+  newparam(:dsc_resource_properties, :array_matching => :all) do
```
Iristyle: Hmm - not sure if `:array_matching => :all` does anything for us here?

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      }.to raise_error(Puppet::Error, /Got nil value for name/)
+    end
+
+    it "should not allow empty" do
+      expect {
+        resource[:name] = ''
+      }.to raise_error(Puppet::ResourceError, /A non-empty name must/)
+    end
+    
+    [ 'value', 'value with spaces', 'UPPER CASE', '0123456789_-', 'With.Period' ].each do |value|
```
Iristyle: Aren't DSC resource naming guidelines stricter than this? I wouldn't expect `<space>` to be allowed?  And I wouldn't expect values starting with numerics?

I'm having a tough time tracking down any specs / docs for the guidelines though - maybe buried somewhere in the CIM spec at https://www.dmtf.org/standards/cim/schemas?

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -0,0 +1,121 @@
...
```
Iristyle: Same here for the spaces... PowerShell 3 language spec says this about `simple-name`:

image

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    it "requires a hash or array of hashes" do
+      expect {
+        resource[:dsc_resource_properties] = "hi"
+      }.to raise_error(Puppet::Error, /dsc_resource_properties should be a Hash/)
+      expect {
+        resource[:dsc_resource_properties] = ["hi"]
+      }.to raise_error(Puppet::Error, /dsc_resource_properties should be a Hash/)
+    end
+  end
+end
```
Iristyle: nit - missing newline

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    it "should not allow empty" do
+      expect {
+        resource[:dsc_resource_properties] = ''
+      }.to raise_error(Puppet::ResourceError, /A non-empty dsc_resource_properties must be specified/)
+    end
+    
+    it "requires a hash or array of hashes" do
+      expect {
+        resource[:dsc_resource_properties] = "hi"
```
Iristyle: Still working this out?

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      }.to raise_error(Puppet::Error, /dsc_resource_properties should be a Hash/)
+      expect {
+        resource[:dsc_resource_properties] = ["hi"]
```
Iristyle: Hmm - what does an array of hashes mean in this context?  Multiple instances of the same resource? That may be complex to implement if that's the intent?

lib/puppet/provider/base_dsc_lite/powershell.rb
jpogran: It's the same now, however I expect it to diverge, and since you want the old way sitting next to the new way, I can't change the format_dsc method

spec/unit/puppet/type/dsc_spec.rb
jpogran: I would think trying to truly validate naming would be a game of catchup we aren't interested in pursuing as we are trying to be loosely typed here, and not get back into the business of validating parts we don't own.

spec/unit/puppet/type/dsc_spec.rb
jpogran: This was part of the exploration for supporting CimInstances, but not fully implemented past being a hash. It was left like this since we ticketed that work in MODULES-6323

spec/unit/puppet/type/dsc_spec.rb
jpogran: This was part of the exploration for supporting CimInstances, but not fully implemented past being a hash. It was left like this since we ticketed that work in MODULES-6323

spec/unit/puppet/type/dsc_spec.rb
jpogran: Same as previous comment

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
jpogran: I had more tests coming in MODULES-6323, or at least I expect whoever implements that ticket will be adding more tests here as we'll be formatting CimInstances

spec/unit/puppet/type/dsc_spec.rb
jpogran: newline added

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
jpogran: newline added

spec/unit/puppet/type/dsc_spec.rb
Iristyle: Looking at your code again, thought you had regexes in place for all of these, but I see there's only one for name.

This is fine, though I don't expect PowerShell class names to be a moving target given they're well defined in grammar specification. Changing that would be backward-incompatible, so I think the rules are pretty set in stone at this point.

spec/unit/puppet/type/dsc_spec.rb
Iristyle: If it's a hash shouldn't it be?

```ruby
resource[:dsc_resource_properties] = { "hi" => "james" }
```

spec/unit/puppet/type/dsc_spec.rb
Iristyle: Probably worth punting to next ticket / PR then

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
Iristyle: Yeah, I would mark these as `private` for now... if you need to add a more complex method for CimInstance formatting later and make that public for the sake of testing it individually, you can do that then.

spec/unit/puppet/type/dsc_spec.rb
Iristyle: Apologies -  I don't know what my last comment was about here - you are rejecting a bare string 🤦‍♂️

PR #16, opened 2018-01-04
(MODULES-6323) EmbeddedInstance support

jpogran:
This commit adds support for EmbeddedInstances (complex types and CIM Instances) when using `Invoke-DscResource` with the generic dsc type.

This commit also updates the `Usage` section of the README to reflect the usage of the dsc type. It does not attempt to update the entire README to reflect the changes from the dsc module, as that is out of scope for this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,14 +11,11 @@
 2. [Prerequisites](#windows-system-prerequisites)
 3. [Setup](#setup)
 4. [Usage](#usage)
-  * [Using DSC Resources with Puppet](#using-dsc-resources-with-puppet)
+  * [Using PSCredential or MSFT_Credential](#using-dsc-resources-with-puppet)
+  * [Using EmbeddedInstance or CimInstance](#using-dsc-resources-with-puppet)
   * [Handling Reboots with DSC](#handling-reboots-with-dsc)
-  * [Installing Packages with DSC](#installing-packages-with-dsc)
-  * [Using Credentials](#using-credentials)
-  * [Setting Registry Values](#setting-registry-values)
-  * [Adding or Removing Windows Features](#adding-or-removing-windows-features)
-  * [Website Installation Example](#website-installation-example)
 5. [Reference](#reference)
+  * [Types and Providers](#types-and-providers)
 6. [Limitations](#limitations)
   * [Known Issues](#known-issues)
   * [Running Puppet and DSC without Administrative Privileges](#running-puppet-and-dsc-without-administrative-privileges)
@@ -52,9 +49,9 @@ See [known issues](#known-issues) for troubleshooting setup.
 
 ## Usage
 
-### Using DSC Resources with Puppet
+### Using PSCredential or MSFT_Credential
 
-You can use a DSC Resource by prefixing each DSC Resource name and parameter with 'dsc_' and lowercasing the values.
+The generic `dsc` type is a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. You can use a DSC Resource by supplying the same properties you would set in a DSC Configuration script inside the `dsc_resource_properties` parameter. For most use cases the `dsc_resource_properties` parameter accepts the same structure as the PowerShell syntax, with the substituion of Puppet syntax for arrays, hashes and other data structures.
 
 So a DSC resource specified in PowerShell...
 
@@ -68,19 +65,121 @@ WindowsFeature IIS {
 ...would look like this in Puppet:
 
 ~~~puppet
-dsc_windowsfeature {'IIS':
-  dsc_ensure => 'present',
-  dsc_name   => 'Web-Server',
+dsc {'iis':
+  dsc_resource_name        => 'WindowsFeature',
+  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_properties  => {
+    ensure => 'present',
+    name   => 'Web-Server',
+  }
 }
 ~~~
 
-All DSC Resource names and parameters have to be in lowercase, for example: `dsc_windowsfeature` or `dsc_name`.
+For the simplest cases, the above example is enough. However there are more advanced use cases in DSC that require more custom syntax in the `dsc` Puppet type. Since the `dsc` Puppet type has no prior knowledge of the type for each property in a DSC Resource, it can't format the hash correctly without some hints.
 
-You can use either `ensure =>` (Puppet's `ensure`) or `dsc_ensure =>` (DSC's `Ensure`) in your manifests for Puppet DSC resource types. If you use both in a Puppet DSC resource, `dsc_ensure` overrides the value in `ensure`, so the value for `ensure` is essentially ignored.
+The `dsc_resource_properties` parameter will recognize any key with a hash value that contains two keys: `dsc_type` and `dsc_properties`, as a indication of how to format the data supplied. The `dsc_type` contains the CimInstance name to use, and the `dsc_properties` contains a hash or an array of hashes representing the data for the CimInstances.
 
-We recommend that you use `dsc_ensure` instead of `ensure`, as it is a closer match for converting the DSC properties to Puppet DSC resources. It also overrides `ensure`, so there is less confusion if both are accidentally included.
+A contrived, but simple example follows:
 
-> Note: While you can use either `ensure =>` (Puppet's `ensure`) or `dsc_ensure =>` (DSC's `Ensure`) in your manifests, there is currently a known issue where `ensure => absent` reports success but does nothing. See [MODULES-2966](https://tickets.puppet.com/browse/MODULES-2966) for details. **Until this issue is resolved, we recommend using `dsc_ensure` exclusively.**
+~~~puppet
+dsc{'foo':
+  dsc_resource_name        => 'xFoo',
+  dsc_resource_module_name => 'xFooBar',
+  dsc_resource_properties  => {
+    ensure  => 'present',
+    fooinfo => {
+      'dsc_type'       => 'FooBarBaz',
+      'dsc_properties' => {
+        "wakka" => "woot",
+        "number"     => 8090
+      }
+    }
+  }
+}
+~~~
+
+### Using PSCredential or MSFT_Credential
+
+Specifying credentials in DSC Resources requires using a PSCredential object. The `dsc` type will automatically create a PSCredential if the `dsc_type` has `MSFT_Credential` as a value.
+
+~~~puppet
+dsc{'foouser':
+  dsc_resource_name        => 'User',
+  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_properties  => {
+    'username'    => 'jane-doe',
+    'description' => 'Jane Doe user',
+    'ensure'      => 'present',
+    'password'    => {
+      'dsc_type' => 'MSFT_Credential',
+      'dsc_properties' => {
+        'user'     => 'jane-doe',
+        'password' => 'StartFooo123&^!'
+      }
+    },
+    'passwordneverexpires' => false,
+    'disabled'             => true,
+  }
+}
+~~~
+
+### Using CimInstance
+
+A DSC Resource may need a more complex type that a simple key value pair, so an EmbeddedInstance is used. An EmbeddedInstance is serialized as CimInstance over the wire. In order to represent a CimInstance in the `dsc` type, we will use the `dsc_type` key to specify which CimInstance to use. If the CimInstance is an array, we append a `[]` to the end of the name.
+
+For example, we'll create a new IIS website using the xWebSite DSC Resource, bound to port 80. We use `dsc_type` to specify a `MSFT_xWebBindingInformation` CimInstance, and append `[]` to indicate that it is an array. Note that we do this even if we are only putting a single value in `dsc_properties`.
+
+~~~puppet
+dsc{'NewWebsite':
+  dsc_resource_name        => 'xWebsite',
+  dsc_resource_module_name => 'xWebAdministration',
+  dsc_resource_properties  => {
+    ensure       => 'Present',
+    state        => 'Started',
+    name         => 'TestSite',
+    physicalpath => 'C:\testsite',
+    bindinginfo  => {
+      'dsc_type'       => 'MSFT_xWebBindingInformation[]',
+      'dsc_properties' => {
+        "protocol" => "HTTP",
+        "port"     => 80
+      }
+    }
+  }
+}
+~~~
+
+To show using more than one value in `dsc_properties`, let's create the same site but now bound to both port 80 and 443.
+
+~~~puppet
+dsc{'NewWebsite':
+  dsc_resource_name        => 'xWebsite',
+  dsc_resource_module_name => 'xWebAdministration',
+  dsc_resource_properties  => {
+    ensure       => 'Present',
+    state        => 'Started',
+    name         => 'TestSite',
+    physicalpath => 'C:\testsite',
+    bindinginfo  => {
+      'dsc_type'       => 'MSFT_xWebBindingInformation[]',
+      'dsc_properties' => [
+        {
+          "protocol" => "HTTP",
+          "port"     => 80
+        },
+        {
+          'protocol'             => 'HTTPS',
+          'port'                 => 443,
+          'certificatethumbprint' => 'F94B4CC4C445703388E418F82D1BBAA6F3E9E512',
+          'certificatestorename'  => 'My',
+          'ipaddress'            => '*'
+        }
+      ]
+    }
+  }
+}
+~~~
+
 
 ### Handling Reboots with DSC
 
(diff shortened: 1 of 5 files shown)


lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
```diff
@@ -38,17 +37,41 @@ def self.format_boolean(value)
...
+          type       = value['dsc_type'].gsub('[]','')
+          properties = value['dsc_properties']
+          output = []
+          
+          properties = properties.is_a?(Hash) ? [properties] : properties
+          properties.each do |p|
+            output << "(New-CimInstance -ClassName '#{type}' -ClientOnly -Property #{format_hash_to_string(p)})"
+          end
+          output = output.join(",\n")
+          "[CimInstance[]]@(\n" + output +  "\n)"
```
Iristyle: I think this is missing the non-array variation of `[CimInstance]` usage.

See https://github.com/puppetlabs/puppetlabs-dsc/blob/master/lib/puppet/provider/templates/dsc/invoke_dsc_resource.ps1.erb#L45-L50

spec/integration/puppet_x/puppetlabs/powershell_hash_formatter_spec.rb
```diff
@@ -13,43 +13,130 @@
...
+        expected = <<-HERE
 @{
 'ensure' = 'Present';
-'bindinginfo' = @{
-'dsc_type' = 'MSFT_xWebBindingInformation[]';
-'dsc_properties' = @{
+'name' = 'foo';
+'state' = 'Started';
+'physicalpath' = 'c:/inetpub/wwwroot';
+'bindinginfo' = [CimInstance[]]@(
```
Iristyle: So for this scenario where the manifest has an EmbeddedInstance that is not a `String[]` (in the MOF schema), but just a single `String` instance like:

```
"dsc_type" => "MSFT_xWebBindingInformation",
"dsc_properties" => {}
```

I would expect that the PowerShell code does not use an array

```powershell
[CimInstance](New-CimInstance -ClassName 'MSFT_xWebBindingInformation' -ClientOnly -Property @{})
```

Which is how it works in DSC module at https://github.com/puppetlabs/puppetlabs-dsc/blob/master/lib/puppet/provider/templates/dsc/invoke_dsc_resource.ps1.erb#L45-L50

The original commit to DSC module has the info on why we did that (using MSFT_xWebsite as the example) - https://github.com/puppetlabs/puppetlabs-dsc/commit/cadf58fbc63c0193501b62f5b7e992a5969b825d

spec/integration/puppet_x/puppetlabs/powershell_hash_formatter_spec.rb
```diff
@@ -13,43 +13,130 @@
...
+      end
+
+      it "should output correct syntax with array CimInstance" do
+        expected = <<-HERE
+@{
+'ensure' = 'Present';
+'name' = 'foo';
+'state' = 'Started';
+'physicalpath' = 'c:/inetpub/wwwroot';
+'bindinginfo' = [CimInstance[]]@(
```
Iristyle: 👍

For `MSFT_xWebBindingInformation[]` this is how I expect the PowerShell code is rendered based on how DSC module works.

Iristyle:
Just a couple of additional comments.

### On `[CimInstance]()` vs `[CimInstance]@()`

In your manifest example for the comment on the PR, you have `MSFT_xWebBindingInformation` in the manifest, and then pass an array in the manifest like `dsc_properties => []`.  The tests are a little different, so maybe that was an early example that changed?  

While we *could* make that syntax work, I think it's probably better to stick with something more explicit like `MSFT_xWebBindingInformation[]`.  While we can infer a `[CimInstance]@()` based on `dsc_properties => []`, we can't always determine if `dsc_properties => {}` is a single instance in an EmbeddedInstance array *or* if it's just a single instance, because it's ambiguous. If we get it wrong, the PowerShell call fails as described in https://github.com/puppetlabs/puppetlabs-dsc/commit/cadf58fbc63c0193501b62f5b7e992a5969b825d.

Alternatively, since it appears you've got `dsc_properties` working as either `[]` or `{}`, we could just use that to infer the right thing (requiring additionally that users always put the `[]` around single instanced arrays so we have a 1:1 mapping for generating the right code).

Any thoughts on which syntax is preferable?  If what I'm trying to convey is confusing, let me know.

### `MSFT_KeyValuePair`

I *think* we still need the special-handling for `MSFT_KeyValuePair`... used in spots like https://github.com/puppetlabs/puppetlabs-dsc/blob/3773d00b7f0cd475ee848bcaee0728d7bcfb03c1/lib/puppet_x/dsc_resources/xPSDesiredStateConfiguration/DSCResources/MSFT_xRemoteFile/MSFT_xRemoteFile.schema.mof#L8

In DSC module, that was added as part of https://github.com/puppetlabs/puppetlabs-dsc/commit/2816ce283f5943272df968ecedb2b746d843164a

Do we want to tackle that in another commit for this ticket or file a new ticket?  It doesn't look like it's used much in actual resources that are MOF based (but maybe more prevalent in class-based resources)?

README.md
```diff
@@ -52,9 +48,9 @@ See [known issues](#known-issues) for troubleshooting setup.
 
 ## Usage
 
-### Using DSC Resources with Puppet
+### Using PSCredential or MSF_Credential
```
michaeltlombardi: Minor fix: MSF**T**_Credential.

README.md
```diff
@@ -52,9 +48,9 @@ See [known issues](#known-issues) for troubleshooting setup.
...
 
-You can use a DSC Resource by prefixing each DSC Resource name and parameter with 'dsc_' and lowercasing the values.
+The generic `dsc` type is a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. You can use a DSC Resource by supplying the same properties you would set in a DSC Configuration script inside the `dsc_resource_properties` parameter. For most use cases the `dsc_resource_properties` parameter accepts the same structure as the PowerShell syntax, with the substituion of Puppet syntax for arrays, hashes and other data structures.
```
michaeltlombardi: May want to add here that you now need to specify the module and resource. It's clear in the example, I think, but could be spelled out in this paragraph.

README.md
```diff
@@ -68,273 +64,133 @@ WindowsFeature IIS {
...
+        "wakka" => "woot",
+        "number"     => 8090
+      }
+    }
+  }
 }
 ~~~
 
-#### Registry Example: Dword and Qword
+### Using PSCredential or MSF_Credential
```
michaeltlombardi: Same nit as above.

README.md
```diff
@@ -68,273 +64,133 @@ WindowsFeature IIS {
...
 }
 ~~~
 
-*Note*: DSC Resources are executed under the SYSTEM context by default, which means you are unable to access any user level Registry key without providing alternate credentials.
-
-### Adding or Removing Windows Features
+### Using CimInstance
 
-You can add or remove Windows Features using Puppet type `dsc_windowsfeature` which maps to the `WindowsFeature` DSC Resource.
+A DSC Resource may need a more complex type that a simple key value pair, so an EmbeddedInstance is used. An EmbeddedInstance are serialized as CimInstance over the wire. In order to represent a CimInstance in the `dsc` type, we will use the `dsc_type` key to specify which CimInstance to use. If the CimInstance is an array, we append a `[]` to the end of the name.
```
michaeltlombardi: > An embeddedInstance ~are~ **is a**...

README.md
```diff
@@ -68,273 +64,133 @@ WindowsFeature IIS {
...
-
-~~~powershell
-[PS]> Get-WindowsFeature
-~~~
 
-### Website Installation Example
+### Handling Reboots with DSC
 
-An end-to-end example installation of a test website.
+Add the following `reboot` resource to your manifest. It must have the name `dsc_reboot` for the `dsc` module to find and use it.
```
michaeltlombardi: Nit: Can we clarify when this takes effect? I can kinda grok it from the example.

jpogran: Out of scope for this PR. This PR only has to do with implementing complex types and documenting them

spec/integration/puppet_x/puppetlabs/powershell_hash_formatter_spec.rb
jpogran: This was solved by relying on the user specifying '[]' in the `dsc_type` key.

jpogran:
It was decided to rely on the user entering in (and keeping synced) two pieces of information when dealing with EmbeddedInstances that can either be a scalar or an array.

The `dsc_type` value must contain a `[]` if the value is an array. This was explained in the updates to the README

README.md
Iristyle: Yeah, let's address the `reboot` situation later.  FWIW, the current solution doesn't work as intended anyhow (i.e. Puppet waits until *all* DSC resources have been evaluated before enacting the reboot, rather than triggering immediately after the DSC resource that requests it - see https://tickets.puppetlabs.com/browse/MODULES-3769).

lib/puppet/type/dsc.rb
```diff
@@ -44,7 +44,11 @@
   end
 
   newparam(:dsc_resource_properties, :array_matching => :all) do
-    desc "DSC Resource Properties"
+    desc <<-HERE
+    The hash of properties to pass to the DSC Resource.
+
+    To express EmbeddedInstances, the `dsc_resource_properties` parameter will reconize any key with a hash value that contains two keys: `dsc_type` and `dsc_properties`, as a indication of how to format the data supplied. The `dsc_type` contains the CimInstance name to use, and the `dsc_properties` contains a hash or an array of hashes representing the data for the CimInstances. If the CimInstance is an array, we append a `[]` to the end of the name.
```
Iristyle: nit: `recognize`

Iristyle: Can we use backticks in inline docs?

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
```diff
@@ -38,17 +37,47 @@ def self.format_boolean(value)
...
+          end
+        end
+
+        def self.format_ciminstance(value)
+          type       = value['dsc_type'].gsub('[]','')
+          properties = value['dsc_properties']
+          output     = []
+          properties = properties.is_a?(Hash) ? [properties] : properties
+
+          properties.each do |p|
```
Iristyle: Can remove `output = []` and shorten this to 

```ruby
output = properties.map do |p|
  "(New-CimInstance -ClassName '#{type}' -ClientOnly -Property #{format_hash_to_string(p)})"
end
```

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
```diff
@@ -38,17 +37,47 @@ def self.format_boolean(value)
...
+              "([PSCustomObject]#{format_hash(value['dsc_properties'])} | new-pscredential)"
+            else
+              format_ciminstance(value)
+            end
+          end
+        end
+
+        def self.format_ciminstance(value)
+          type       = value['dsc_type'].gsub('[]','')
+          properties = value['dsc_properties']
```
Iristyle: Probably not as clear, but a common Ruby trick to always get an array here is:

```ruby
properties = [value['dsc_properties']].flatten
```

i.e.

```
[15] pry(main)> [{}].flatten
=> [{}]
[16] pry(main)> [[{}, {}]].flatten
=> [{}, {}]
```

spec/integration/puppet_x/puppetlabs/powershell_hash_formatter_spec.rb
Iristyle: All these examples look good now.

spec/integration/puppet_x/puppetlabs/powershell_hash_formatter_spec.rb
```diff
@@ -13,44 +13,189 @@
...
+        })
+        expect(result).to eq expected.strip
+      end
+
+      it "should output correct syntax with MSFT_KeyValuePair" do
+        expected = <<-HERE
+@{
+'destinationPath' = 'c:\fileName.jpg';
+'uri' = 'http://www.contoso.com/image.jpg';
+'userAgent' = '[Microsoft.PowerShell.Commands.PSUserAgent]::InternetExplorer';
```
Iristyle: Was this intended to be the literal string `'[Microsoft.PowerShell.Commands.PSUserAgent]::InternetExplorer'` or was this intended to return the constant from the .NET class?

jpogran: It has to be quoted in the Puppet manifest, or it breaks the parser. Since it's a string in the Puppet manifest, it's parsed as a string and set as quoted in the PowerShell hash. Technically the PowerShell hash could take a type or a string as parsable, however it's irrelevant as PowerShell will cast whatever is passed to the property based on the parameter declaration of the DSC Resource. In this case, it's typed as a string and passed in a PSBoundParameters block directly to Invoke-WebRequest, which has UserAgent as a string parameter.

Even if we can't rely on the type casting PowerShell does under the covers, I don't see  this being an issue in the future, for if it needs to be a 'complex type' then it needs to be an EmbeddedInstance class declaration so it can be serialized into a CimInstance.

jpogran:
@michaeltlombardi I've addressed some of your notes about the readme by adding more to the types section and fleshing out some of the builtin type desc blocks.

spec/integration/puppet_x/puppetlabs/powershell_hash_formatter_spec.rb
Iristyle: Yeah, I understood the manifest -> generated code mechanics, but didn't connect the dots to the PS conversion.

So this ends up being a pretty nifty trick that we can rely on PS conversion for stringified value -> actual .NET type. 

This is good to know / something we should probably consider documenting for users.

https://blogs.msdn.microsoft.com/powershell/2013/06/11/understanding-powershells-type-conversion-magic/

README.md
```diff
@@ -68,282 +65,162 @@ WindowsFeature IIS {
...
-  dsc_hash_parameter => {
-    'key1' => 'value1',
-    'key2' => 'value2'
-  },
-}
-~~~
+### Using CimInstance
 
-### Using Credentials
+A DSC Resource may need a more complex type that a simple key value pair, so an EmbeddedInstance is used. An EmbeddedInstance is serialized as CimInstance over the wire. In order to represent a CimInstance in the `dsc` type, we will use the `dsc_type` key to specify which CimInstance to use. If the CimInstance is an array, we append a `[]` to the end of the name.
```
Iristyle: minor - `may need a more complex type than`

README.md
```diff
@@ -68,282 +65,162 @@ WindowsFeature IIS {
...
-### Handling Reboots with DSC
-
-Add the following `reboot` resource to your manifest. It must have the name `dsc_reboot` for the `dsc` module to find and use it.
+A contrived, but simple example follows:
 
 ~~~puppet
-reboot { 'dsc_reboot' :
-  message => 'DSC has requested a reboot',
-  when => 'pending'
+dsc{'foo':
```
Iristyle: Minor formatting - other examples use `dsc {`, not `dsc{`

README.md
```diff
@@ -68,282 +65,162 @@ WindowsFeature IIS {
...
-
-    $record = $View.GetType().InvokeMember("Fetch", "InvokeMethod", $Null, $View, $Null)
-    $property = $record.GetType().InvokeMember("StringData", "GetProperty", $Null, $record, 1)
-
-    $View.GetType().InvokeMember("Close", "InvokeMethod", $Null, $View, $Null)
-
-    return $property
-  }catch{
-    throw "Failed to read MSI file. The error was: {0}." -f $_
+dsc{'foouser':
```
Iristyle: Minor formatting - other examples use `dsc {`, not `dsc{`

README.md
```diff
@@ -68,282 +65,162 @@ WindowsFeature IIS {
...
-  dsc_username             => 'jane-doe',
-  dsc_description          => 'Jane Doe user',
-  dsc_ensure               => present,
-  dsc_password             => {
-    'user' => 'jane-doe',
-    'password' => 'jane-password'
-  },
-  dsc_passwordneverexpires => false,
-  dsc_disabled             => true,
+dsc{'NewWebsite':
```
Iristyle: Minor formatting - other examples use `dsc {`, not `dsc{`

README.md
```diff
@@ -68,282 +65,162 @@ WindowsFeature IIS {
...
-
-You can use either `ensure =>` (Puppet's `ensure`) or `dsc_ensure =>` (DSC's `Ensure`) in your manifests for Puppet DSC resource types. If you use both in a Puppet DSC resource, `dsc_ensure` overrides the value in `ensure`, so the value for `ensure` is essentially ignored.
+For the simplest cases, the above example is enough. However there are more advanced use cases in DSC that require more custom syntax in the `dsc` Puppet type. Since the `dsc` Puppet type has no prior knowledge of the type for each property in a DSC Resource, it can't format the hash correctly without some hints.
 
-We recommend that you use `dsc_ensure` instead of `ensure`, as it is a closer match for converting the DSC properties to Puppet DSC resources. It also overrides `ensure`, so there is less confusion if both are accidentally included.
+The `dsc_resource_properties` parameter will recognize any key with a hash value that contains two keys: `dsc_type` and `dsc_properties`, as a indication of how to format the data supplied. The `dsc_type` contains the CimInstance name to use, and the `dsc_properties` contains a hash or an array of hashes representing the data for the CimInstances.
```
Iristyle: We should probably link out to / explain a bit more about what a `CimInstance` is... for a future update to README

lib/puppet/type/dsc.rb
```diff
@@ -44,7 +44,11 @@
   end
 
   newparam(:dsc_resource_properties, :array_matching => :all) do
-    desc "DSC Resource Properties"
+    desc <<-HERE
+    The hash of properties to pass to the DSC Resource.
+
+    To express EmbeddedInstances, the dsc_resource_properties parameter will reconize any key with a hash value that contains two keys: dsc_type and dsc_properties, as a indication of how to format the data supplied. The dsc_type contains the CimInstance name to use, and the dsc_properties contains a hash or an array of hashes representing the data for the CimInstances. If the CimInstance is an array, we append a [] to the end of the name.
```
Iristyle: Still says 'reconize` instead of `recognize`

Iristyle:
We can address the minor typos, etc in #19 I think?

PR #17, opened 2018-01-08
(MODULES-6132) Rewrite multiple module version Beaker tests

Iristyle:
Rewrites two old remaining Beaker tests / adds a new test fixture DSC resource / removes dead code:

 - Previously there were 2 existing tests that verified load behavior
   of DSC resources:

   * duplicate_different_module_version - this would create 2
     installed copies of the same resource, re-versioning and
     poisioning the second so that it was known in tests if it was
     loaded accidentally instead of the desired copy

     This test relied on an implementation detail within generated
     types which no longer makes sense, and as such is deleted

  * duplicate_same_module_version - similarly, this module would copy
    an existing DSC resource from the vendored module path into the
    users part of $ENV:PSModulePath

    This test existed simply to demonstrate that a DSC module in the
    PowerShell module path would not interfere with loading a module
    by explicit path

    This test relies on the same implementation detail as above / can
    be replicated in a different way.

 - Modify the dsc_utils "install_fake_reboot_resource" helper to
   additionally copy PuppetFakeResource2 during tests. These files
   will remain latent until a test opts into consuming them. They can
   be  referred to by explicit path, but to make them useful in tests
   they will generally be copied to $ENV:PSModulePath, renaming
   PuppetFakeResource2 -> PuppetFakeResource in the process so that
   DSC runtime may find them by name / version.

 - Add new test `multiple_dsc_resources_in_PSModulePath` that performs
   similar checks against the new `dsc` type.

   This test:

   * Copies PuppetFakeResource -> $PSHome part of module path
   * Copies PuppetFakeResource2 (as PuppetFakeResource) ->
     $ENV:ProgramFiles part of module path
   * Verifies DSC sees both resources
   * Uses the ambiguous `ModuleName` reference `PuppetFakeResource` to
     try to apply a manifest and fails with expected DSC error as DSC
     cannot determine which module to load
   * Demonstrates future behavior of loading by a specific version of
     the resource by version number in a manifest, that will be
     implemented in MODULES-5845

Also adds a new test:

 - With the duplicate_same_module_version test removed, add a test
   that demonstrates loading / applying a custom user module from
   the $PSHome module path (from $ENV:PSModulePath)

   This simply demonstrates that dsc_resource_module_name may be a
   name of an available resource in addition to the previously tested
   fully qualified path.

   NOTE: When copying files from the vendored resource path, they
   must have permissions reset when copying back out to paths in
   $ENV:PSModulePath, else they will not load under DSC.
af1ccb3


Iristyle:
Verified locally - running through adhoc at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/All/job/forge-windows_puppetlabs-dsc_lite_init-manual-parameters_adhoc/4/

Iristyle:
I hadn't yet tested the full suite and it looks like one of the other tests broke as a result in CI - looking into it.

tests/integration/tests/dsc_type/multiple_dsc_resources_in_PSModulePath.rb
```diff
@@ -0,0 +1,141 @@
...
+inject_site_pp(master, get_site_pp_path(master), create_site_pp(master))
+
+confine_block(:to, :platform => 'windows') do
+  step 'Sync DSC resource implementations to agents'
+  on(agents, puppet('agent -t --environment production'), :acceptable_exit_codes => [0,2])
+
+  step 'Copy PuppetFakeResource implementations to system PSModulePath locations'
+  installed_path = '/cygdrive/c/ProgramData/PuppetLabs/puppet/cache/lib/puppet_x/dsc_resources'
+
+  # put PuppetFakeResource v1 in $PSHome\Modules
```
Iristyle: The `Install-Module` like variant introduces a directory structure like:

```
$PSHome\Modules\PuppetFakeResource\1.0\PuppetFakeResource
$PSHome\Modules\PuppetFakeResource\2.0\PuppetFakeResource
```

In my testing, the behavior was identical in that DSC presents with `Resource PuppetFakeResource was not found`.

We should probably still add a test that behaves like `Install-Module` as users are probably more likely to install resources that way, rather than the xcopy style presented here.

Iristyle:
Ad-hoc passed through https://jenkins-master-prod-1.delivery.puppetlabs.net/view/All/job/forge-windows_puppetlabs-dsc_lite_i18n-ruby_adhoc/3/

tests/files/dsc_puppetfakeresource/PuppetFakeResource2/PuppetFakeResource.psd1
```diff
@@ -0,0 +1,120 @@
...
+GUID = '69a082d2-80c8-46fb-bb8d-b102010ff2e9'
+
+# Author of this module
+Author = 'vagrant'
+
+# Company or vendor of this module
+CompanyName = 'Unknown'
+
+# Copyright statement for this module
+Copyright = '(c) 2015 vagrant. All rights reserved.'
```
glennsarti: This may be a concern.  Should either change to Puppet Inc. or remove the copyright notice.

tests/files/dsc_puppetfakeresource/PuppetFakeResource2/PuppetFakeResource.psd1
```diff
@@ -0,0 +1,120 @@
...
+@{
+
+# Script module or binary module file associated with this manifest.
+# RootModule = ''
+
+# Version number of this module.
+ModuleVersion = '2.0'
+
+# ID used to uniquely identify this module
+GUID = '69a082d2-80c8-46fb-bb8d-b102010ff2e9'
```
glennsarti: This is the same GUID as the V1 resource.  Is this intentional?

glennsarti:
Typo in commit https://github.com/puppetlabs/puppetlabs-dsc_lite/pull/17/commits/a2d132b3550f36752ddf52123de06d1f0e725049

`"install_fake_reook_resource"` should be `"install_fake_reboot_resource"`

glennsarti:
Changes look legit and adhoc passed.

tests/files/dsc_puppetfakeresource/PuppetFakeResource2/PuppetFakeResource.psd1
Iristyle: I asked the same question when I copied it over... I couldn't find any real DSC resources where GUID changed from one version to the next, so I left it the same.

Compare https://github.com/PowerShell/PSDscResources/blob/2.0.0.0-PSGallery/PSDscResources.psd1#L13 to https://github.com/PowerShell/PSDscResources/blob/2.8.0.0-PSGallery/PSDscResources.psd1#L13 (spoiler alert, GUID is the same)

tests/files/dsc_puppetfakeresource/PuppetFakeResource2/PuppetFakeResource.psd1
Iristyle: Yeah, I can add another commit to do that. That's from the original PSD1 added at https://github.com/puppetlabs/puppetlabs-dsc/blob/02298740e988643af50142edbdf77cd102ff6c3a/tests/files/reboot/PuppetFakeResource/PuppetFakeResource.psd1

I thought it was @jpogran that initially created the resource, but looks like @cowofevil added it to the repo in https://github.com/puppetlabs/puppetlabs-dsc/pull/156

In any event, we'll get that removed (should probably PR DSC repo as well)

jpogran: I created it for manual testing, but @cowofevil added it to the acceptance testing.

tests/files/dsc_puppetfakeresource/PuppetFakeResource2/PuppetFakeResource.psd1
jpogran: GUID are not supposed to change, and are a carryover from the PowerShell module manifest semantics they inherit.

The truly 'unique' identifier is the namespace from the name of the module and the version.

tests/integration/tests/dsc_type/multiple_dsc_resources_in_PSModulePath.rb
jpogran: I think we should be using the versioning folder path structure that the PackageManagement tooling uses, as that is the way a user will be expecting to use this.

tests/files/dsc_puppetfakeresource/PuppetFakeResource2/PuppetFakeResource.psd1
glennsarti: Ahh ok....so the U is not for unique :-)  Thanks MS

tests/integration/tests/dsc_type/multiple_dsc_resources_in_PSModulePath.rb
Iristyle: Yeah, I think we should do that in addition to what I did here as they're both valid ways of installing PS modules.  For instance, users could keep DSC resources in their Puppet Code Manager control repo and use Puppet to lay down files rather than using upstream Gallery / hosting a local share / etc given it has fewer moving parts (not saying this is a good idea, but a plausible distro mechanism that's pretty easy to understand).

I think it would also be preferable to have some fixtures we can actually call `Install-Module` on, rather than manually rigging up the directories, if it's not too much of a pain to zip things up / point to a local repo (should work via UNC share?).

Let's plan on taking a stab at that for next batch of test changes.

Iristyle:
```
Test Suite: tests @ 2018-01-09 11:27:25 -0800

      - Host Configuration Summary -


              - Test Case Summary for suite 'tests' -
       Total Suite Time: 508.39 seconds
      Average Test Time: 24.21 seconds
              Attempted: 21
                 Passed: 19
                 Failed: 0
                Errored: 2
                Skipped: 0
                Pending: 0
                  Total: 21
```

2 failures were files that I had backups of locally that were run accidentally, ignore those ;0

Iristyle:
Yeah, I'm not in ❤️  with the tests either, but... baby steps as we move toward the overhaul.

PR #18, opened 2018-01-08
(MODULES-5847) Clarify tradeoffs

michaeltlombardi:
Prior to this commit there is no information captured
in the repository for _why_ a user might (and might
not) want to switch from `DSC` to `DSC_Lite`. This
commit adds that information for potential users.

### DIFF
--- a/README_Tradeoffs.md
+++ b/README_Tradeoffs.md
@@ -1,0 +1,134 @@
+# DSC implementations: tradeoffs
+`dsc_lite` is a new approach to managing nodes by leveraging Microsoft's [Desired State Configuration](https://docs.microsoft.com/en-us/powershell/dsc/overview).
+Puppet has an existing approach in the form of the supported [`dsc` Module](https://github.com/puppetlabs/puppetlabs-dsc).
+
+This document explains the benefits and drawbacks of each approach and why a second approach has been implemented.
+
+## Overview
+The approach to managing DSC Resources in `dsc_lite` is to use a generalized Puppet call.
+This lets you use any DSC Resource available on the target node but requires you to set all of the configuration data correctly.
+
+Instead of writing a declaration in your Puppet manifest that looks like this:
+
+```puppet
+dsc_file {'fruit_file':
+  dsc_ensure          => 'present',
+  dsc_type            => 'File',
+  dsc_destinationpath => 'C:\\Fruit.txt',
+  dsc_contents        => 'Apple, Banana, Cherry.',
+}
+```
+
+You'll write a simplified declaration that looks like this:
+
+```puppet
+dsc {'fruit_file':
+  dsc_resource_name => 'File',
+  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_properties => {
+    ensure          => 'present',
+    content         => 'Apple, Banana, Cherry.',
+    destinationpath => 'C:\\Fruit.txt',
+  }
+}
+```
+
+The latter is a generalized resource that allows you to use _any_ DSC Resource which is compatible with [`Invoke-DscResource`](https://docs.microsoft.com/en-us/powershell/module/psdesiredstateconfiguration/invoke-dscresource?view=powershell-5.1) and is available on the machine that you are configuring, while the former only allows you to use the DSC Resources bundled with the Puppet `dsc` module.
+
+## Why should you use the new `dsc_lite` module?
+
+Use `dsc_lite` if one or more of the following apply to you:
+
++ You need to use one or more DSC Resources which are _not_ [vendored](#vendoring) into Puppet's `dsc` module
++ You want to use class-based or custom DSC Resources
++ You need to use multiple versions of the same DSC resource
++ You want pluginsync to benefit from a [reduced footprint](#pluginsync-impact) - the file count for the module is reduced from thousands of files to a handful
++ You don't want to build your own custom version of the Puppet `dsc` module to include custom resources
++ You can live without the [validation](#validation) of the existing Puppet `dsc` module
++ You don't mind managing the installation of PowerShell modules containing DSC Resources on your nodes
++ You want to use the latest versions of DSC Resources even if they're not [vendored](#vendoring)
+
+## Why should you keep using the Puppet `dsc` module?
+
+You should consider continuing to use the existing Puppet `dsc` module if one or more of the following apply to you:
+
++ You want [earlier feedback about the validity of manifest code](#validation)
++ You're only using Microsoft-provided DSC Resources - ie, not using custom, community-maintained, or class-based DSC Resources
++ You've already got a good [workflow for building an internal release of the Puppet `dsc` module with custom resources](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/README_BUILD.md)
++ You don't want to manage installing DSC Resources on your nodes
++ You need the granularity of reporting in the Puppet Enterprise (PE) console which is based on each resource type being different - allowing you to filter and group on DSC Resources
+
+## Additional considerations
+
+The `dsc_lite` module is currently in an early, unsupported state and undergoing rapid iteration.
+It is intended to be a lighter weight alternative to the existing module which primarily trades off flexibility for ease-of-use and safety.
+
+Switching between the two approaches will require some work to update the resource declarations in your manifests.
+
+A primary drawback of using `dsc_lite` is that you're responsible for installing DSC Resources onto all of the nodes you wish to configure with it - the module no longer vendors the DSC Resources for you.
+
+### Vendoring
+
+The existing Puppet `dsc` module vendors numerous DSC Resources into it - that is, the Puppet `dsc` module includes [many PowerShell DSC Resources](https://github.com/puppetlabs/puppetlabs-dsc/tree/master/lib/puppet_x/dsc_resources) inside it.
+This means that those DSC Resources are automatically on every machine which has the Puppet `dsc` module without you having to manage them individually.
+However, this means that if there are DSC Resources you do _not_ want to sync, or DSC Resources not vendored in the Puppet `dsc` module, you'll need to [rebuild the module to your own specifications](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/README_BUILD.md).
+
+The `dsc_lite` module does not vendor the DSC Resources for you, but neither does it limit you to only using vendored DSC Resources.
+Any DSC Resource available on the machine can be used by the `dsc_lite` module.
+
+#### Pluginsync impact
+Because the Puppet `dsc` module vendors many different DSC Resources it also has a large footprint and includes thousands of files.
+Because pluginsync downloads files from _every_ module in the agent's environment modulepath, regardless of whether the node uses them, this means that every machine using an environment that includes the Puppet `dsc` module will get these thousands of files synced to it.
+This can be very slow and has the most noticeable impact amongst short-lived machines.
+
+### Validation
+The existing Puppet `dsc` module also gives you valuable development and compile-time validation.
+The Puppet `dsc` module is aware of the types required for the DSC Resources, including whether a property should be an `int` or a `string`, whether it should be a well-formed path, and whether or not the property is required for the DSC Resource to properly run.
+Because of this awareness, the Puppet `dsc` module is able to give you immediate feedback when parsing your manifest - letting you know right away if you've made a syntactic mistake - and useful, human readable feedback at compile time.
+
+For example, if we specify an integer for a property that requires a path:
+
+```puppet
+dsc_file {'fruit_file':
+  dsc_ensure          => 'present',
+  dsc_type            => 'File',
+  dsc_destinationpath => 1,
+}
+```
+
+In the Puppet `dsc` module we get a well-formed error and the invocation never runs on the machine.
+
+```txt
+Notice: Compiled catalog for win-tmihi4gdjn2.localdomain in environment production in 0.30 seconds
+Error: Parameter dsc_destinationpath failed on Dsc_file[fruit_file]: Invalid value '1'. Should be a string at C:/code/old.pp:1
+```
+
+If we do the same with the new `dsc_lite` module:
+
+```puppet
+dsc {'fruit_file':
+  dsc_resource_name => 'File',
+  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_properties => {
+    ensure          => 'present',
+    destinationpath => 1,
+  }
+}
+```
+
+We get a less clear error message from the code _actually_ running (and erroring) on the machine.
+
+```txt
+Notice: Compiled catalog for win-tmihi4gdjn2.localdomain in environment production in 0.46 seconds
+Error: /Stage[main]/Main/Dsc[fruit_file]: Could not evaluate: Convert property 'destinationpath' value from type 'SINT64' to type 'STRING' failed
+ At line:9, char:2
+ Buffer:
+irectResourceAccess";
+};^
+
+insta
+
+Notice: Applied catalog in 3.33 seconds
+```
+
+Because `dsc_lite` is unaware of the types and requirements of any given DSC Resource, the module can't provide validation without actually calling the code.


README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Why Should You Keep Using the DSC Module?
+
+You should consider continuing to use the existing DSC module if one or more of the following apply to you:
+
++ You want the parse-time validation and safety of the DSC Module
++ You're not using custom, composite, or class-based DSC resources
++ You've already got a good [workflow for building an internal release of the DSC module with custom resources](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/README_BUILD.md)
++ You don't want to manage installing PowerShell modules on your nodes
++ You need the granularity of reporting in the PE console which is based on each resource type being different - allowing you to filter and group on dsc resources
```
clairecadman: *DSC

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+You should use this module if one or more of the following apply to you:
+
++ You need to use one or more DSC resources which are _not_ vendored into our DSC module
++ You want to use class-based or composite DSC resources
++ You need to use multiple versions of the same DSC resource
++ You want a reduced pluginsync impact
++ You don't want to build your own custom version of the DSC module to include custom resources
++ You can live without the safety and guardrails of the existing DSC module
++ You don't mind managing getting PowerShell modules containing DSC resources to your nodes
```
clairecadman: managing OR getting?

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
++ You need the granularity of reporting in the PE console which is based on each resource type being different - allowing you to filter and group on dsc resources
+
+## Additional Considerations
+
+The `DSC_Lite` module is currently in an early, unsupported state and undergoing rapid iteration.
+It is intended to be a lighter weight alternative to the existing module which primarily trades off flexibility for ease-of-use and safety.
+
+Switching between the two approaches will require some work to update the resource declarations in your manifests.
+
+A primary drawback of this module is that you're now responsible for managing getting PowerShell modules onto all of the nodes you wish to configure with this module - it no longer vendors the DSC resources for you.
```
clairecadman: managing OR getting?

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
+# DSC Implementations: Tradeoffs
+This module, DSC_Lite, represents a new approach to managing nodes by leveraging Microsoft's [Desired State Configuration](https://docs.microsoft.com/en-us/powershell/dsc/overview).
```
clairecadman: Suggested wording change:

DSC_Lite is a new approach to managing nodes....etc (no "this module").

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+We have an existing approach in the form of our supported [DSC Module](https://github.com/puppetlabs/puppetlabs-dsc).
+
+This document aims primarily to explain the pros and cons of each approach in comparison to each other, and to explain _why_ a second approach has been implemented.
```
clairecadman: Suggested wording change:

"This document explains the pros and cons of each approach, and _why_ a DSC_Lite has been implemented."

I haven't seen _why_ like this before, but assume is means something written like this?

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Overview
+The approach to managing DSC resources in this module is to use a generalized puppet call which lets you use any DSC resource available on the target node but requires you to get all of the configuration data correct.
```
clairecadman: Suggested wording change:

"The approach to managing resources in DSC_Lite is to use a generalized Puppet call. This lets you use any DSC resource available on the target node but also requires you to get all of the configuration data correctly".

Capitalize Puppet.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+Instead of using resources in your puppet manifest that look like this:
```
clairecadman: Capitalize Puppet.

README_Tradeoffs.md
michaeltlombardi: Both, you're managing the retrieval of those modules. This should be reworded for clarity though.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_properties => {
+    ensure          => 'present',
+    content         => 'Apple, Banana, Cherry.',
+    destinationpath => 'C:\\Fruit.txt',
+  }
+}
+```
+
+The latter is a generalized resource that allows you to use _any_ DSC module / resource available on the machine you are configuring.
```
clairecadman: "..on the machine that you are configuring." (add "that")

README_Tradeoffs.md
michaeltlombardi: As above, this needs rewording.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+Let's take a look at why you might want to be using this new module.
```
clairecadman: Suggest omitting sentence. Says it in the title below.

README_Tradeoffs.md
michaeltlombardi: The emphasis is a habit of mine, and possibly not a useful one in this context.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Why Should You Use the New DSC_Lite Module?
+
+You should use this module if one or more of the following apply to you:
```
clairecadman: Suggestion: omit "you should"

michaeltlombardi:
@clairecadman, there was earlier discussion as to whether this document should itself be rolled into the README. Before merge we'll need to figure that out. What're your thoughts?

README_Tradeoffs.md
clairecadman: DSC_Lite, instead of "this module".

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
++ You need to use multiple versions of the same DSC resource
++ You want a reduced pluginsync impact
++ You don't want to build your own custom version of the DSC module to include custom resources
++ You can live without the safety and guardrails of the existing DSC module
++ You don't mind managing getting PowerShell modules containing DSC resources to your nodes
++ You want to use the latest versions of DSC resources even they're not vendored
+
+## Why Should You Keep Using the DSC Module?
+
+You should consider continuing to use the existing DSC module if one or more of the following apply to you:
```
clairecadman: Maybe omit "consider"

"You should continue to use..."

README_Tradeoffs.md
clairecadman: Spell out PE when using for the first time in the doc - "Puppet Enterprise (PE)".

README_Tradeoffs.md
michaeltlombardi: The reason for using the language here is to convey that neither approach is perfect and most users likely have a mix of desires from both lists.

Is there a better way to word this sentiment? If we leave this in a tradeoff document, does that even need to be spelled out?

README_Tradeoffs.md
clairecadman: Suggested reword:

"A primary drawback of DSC_Lite is that you're responsible for managing PowerShell modules on all the nodes you want to configure, and it no longer vendors the DSC resources for you". Unless this is the same thing? Then maybe say one or the other?

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+    destinationpath => 'C:\\Fruit.txt',
+  }
+}
+```
+
+The latter is a generalized resource that allows you to use _any_ DSC module / resource available on the machine you are configuring.
+Let's take a look at why you might want to be using this new module.
+
+## Why Should You Use the New DSC_Lite Module?
+
```
clairecadman: We don't usually capitalize titles. e.g. it would be:

## Why should you use the new DSC_Lite module?

Same for all the other titles.

README_Tradeoffs.md
michaeltlombardi: The phrase `managing getting` is definitely too ambiguous and needs to be replaced. Should we replace it with the same phrase throughout the document?

To me, the sentences read to have _almost_ the same meaning, except for a shade of extra responsibility in your wording. I think that might be useful though, to emphasize that it's an additional burden on the user.

README_Tradeoffs.md
michaeltlombardi: What about the top-level title? Should it switch to

# DSC implementations: tradeoffs

Or stay as-is?

clairecadman: Yes I would switch to that.

README_Tradeoffs.md
clairecadman: Yeah we don't usually use use italics or bolding in the docs.. but as this won't be in the docs I think its fine if you think it is useful? Either way!

README_Tradeoffs.md
clairecadman: Yeah maybe saying "managing the retrieval...". Though maybe that is a bit long. Whichever you think the users of this would understand, but stay consistent throughout.

README_Tradeoffs.md
michaeltlombardi: Okay, excellent.

README_Tradeoffs.md
michaeltlombardi: Dropping it for now, I don't think it adds enough to break from convention.

README_Tradeoffs.md
clairecadman: Ah okay. Maybe:

"However, if one or more of the following statements apply to you, consider continuing to use the existing DSC module:..." 

"Consider continuing..." doesn't read that well, but I can't think of a good alternative. You could alway omit the sentence all together, same with the first sentence in the section above, as the title says it all. Or I think fine to keep as it is too.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
++ You want to use the latest versions of DSC resources even they're not vendored
+
+## Why Should You Keep Using the DSC Module?
+
+You should consider continuing to use the existing DSC module if one or more of the following apply to you:
+
++ You want the parse-time validation and safety of the DSC Module
++ You're not using custom, composite, or class-based DSC resources
++ You've already got a good [workflow for building an internal release of the DSC module with custom resources](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/README_BUILD.md)
++ You don't want to manage installing PowerShell modules on your nodes
```
michaeltlombardi: @clairecadman, here the language we use is `manage installing` - is this acceptable, or should this be updated along with the `managing getting` phrase?

clairecadman: If it means the same I would change it to what the others are, or have them all say "manage installing". If it is different I think "manage installing" is okay.

clairecadman:
@michaeltlombardi I think some info from this would be useful in a README, but only what they need to know.

README_Tradeoffs.md
michaeltlombardi: I've updated all of the language to reflect that users need to take responsibility for installing the PowerShell modules.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
+# DSC implementations: tradeoffs
+DSC_Lite is a new approach to managing nodes by leveraging Microsoft's [Desired State Configuration](https://docs.microsoft.com/en-us/powershell/dsc/overview).
+We have an existing approach in the form of our supported [DSC Module](https://github.com/puppetlabs/puppetlabs-dsc).
```
glennsarti: nit:  Is this "We" as in the reader + author or "We" as in Puppet Inc.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+This document explains the pros and cons of each approach and why a second approach has been implemented.
```
glennsarti: nit: Use explicit meaning not abbreviations   ... the benefits and disadvantages

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+The latter is a generalized resource that allows you to use _any_ DSC module / resource available on the machine that you are configuring.
+
+## Why should you use the new DSC_Lite module?
+
+Use DSC_Lite if one or more of the following apply to you:
+
++ You need to use one or more DSC resources which are _not_ vendored into our DSC module
++ You want to use class-based or composite DSC resources
++ You need to use multiple versions of the same DSC resource
++ You want a reduced pluginsync impact
```
glennsarti: Does this assume readers know there is an impact?

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
++ You don't want to build your own custom version of the DSC module to include custom resources
++ You can live without the safety and guardrails of the existing DSC module
```
glennsarti: What are the guard rails?  Hasn't been explained yet

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
++ You don't mind managing the installation of PowerShell modules containing DSC resources on your nodes
++ You want to use the latest versions of DSC resources even they're not vendored
```
glennsarti: What is vendoring?  Hasn't been explained yet

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Why should you keep using the Puppet DSC module?
+
+You should consider continuing to use the existing Puppet DSC module if one or more of the following apply to you:
+
++ You want the parse-time validation and safety of the DSC Module
++ You're not using custom, composite, or class-based DSC resources
```
glennsarti: Doesn't this exclude everything?  Perhaps specify what you are using e.g.

You are only using the Microsoft provided DSC resources

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
++ You've already got a good [workflow for building an internal release of the DSC module with custom resources](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/README_BUILD.md)
++ You don't want to manage installing PowerShell modules on your nodes
++ You need the granularity of reporting in the Puppet Enterprise (PE) console which is based on each resource type being different - allowing you to filter and group on DSC resources
+
+## Additional considerations
```
glennsarti: This seems like it should be above the '## Why should you use the new DSC_Lite module?' section'

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
+# DSC implementations: tradeoffs
+DSC_Lite is a new approach to managing nodes by leveraging Microsoft's [Desired State Configuration](https://docs.microsoft.com/en-us/powershell/dsc/overview).
+We have an existing approach in the form of our supported [DSC Module](https://github.com/puppetlabs/puppetlabs-dsc).
+
+This document explains the pros and cons of each approach and why a second approach has been implemented.
+
+## Overview
+The approach to managing DSC resources in DSC_Lite is to use a generalized Puppet call.
```
jpogran: It's `DSC Resource`, not `DSC resource`. Think `Desired State Configuration Resource`. Apply to any variation such as `DSC Resources`, etc.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+This lets you use any DSC resource available on the target node but requires you to set all of the configuration data correctly.
+
+Instead of using resources in your Puppet manifest that look like this:
```
jpogran: Instead of writing a declaration in your Puppet manifest that looks like this:

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+```puppet
+dsc_file {'fruit_file':
+  dsc_ensure          => 'present',
+  dsc_type            => 'File',
+  dsc_destinationpath => 'C:\\Fruit.txt',
+  dsc_contents        => 'Apple, Banana, Cherry.',
+}
+```
+
+You'll use resources that look like this:
```
jpogran: You'll write a simplified declaration that looks like this:

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_properties => {
+    ensure          => 'present',
+    content         => 'Apple, Banana, Cherry.',
+    destinationpath => 'C:\\Fruit.txt',
+  }
+}
+```
+
+The latter is a generalized resource that allows you to use _any_ DSC module / resource available on the machine that you are configuring.
```
jpogran: ,while the former only allows you to use the DSC Resources bundled with the dsc module

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Why should you use the new DSC_Lite module?
+
+Use DSC_Lite if one or more of the following apply to you:
+
++ You need to use one or more DSC resources which are _not_ vendored into our DSC module
++ You want to use class-based or composite DSC resources
```
jpogran: Canna use composite in either case (https://docs.microsoft.com/en-us/powershell/dsc/directcallresource)

README_Tradeoffs.md
jpogran: I agree with @glennsarti . Some of these points deserve more than a bullet point, as they are the purpose we are making this. You can look at the original design docs for more info

README_Tradeoffs.md
jpogran: It's not PowerShell modules, it's `DSC Resources`.

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
...
++ You need the granularity of reporting in the Puppet Enterprise (PE) console which is based on each resource type being different - allowing you to filter and group on DSC resources
+
+## Additional considerations
+
+The `DSC_Lite` module is currently in an early, unsupported state and undergoing rapid iteration.
+It is intended to be a lighter weight alternative to the existing module which primarily trades off flexibility for ease-of-use and safety.
+
+Switching between the two approaches will require some work to update the resource declarations in your manifests.
+
+A primary drawback of using DSC_Lite is that you're responsible for installing PowerShell modules onto all of the nodes you wish to configure with it - the module no longer vendors the DSC resources for you.
```
jpogran: These aren't powershell modules they are installing, its `DSC Resources`

README_Tradeoffs.md
```diff
@@ -0,0 +1,68 @@
+# DSC implementations: tradeoffs
+DSC_Lite is a new approach to managing nodes by leveraging Microsoft's [Desired State Configuration](https://docs.microsoft.com/en-us/powershell/dsc/overview).
```
jpogran: Don't we use lower case for our module names? For example, 'The Puppet dsc_lite module performs the following:'

README_Tradeoffs.md
jpogran: Puppet has

README_Tradeoffs.md
michaeltlombardi: @clairecadman, is this accurate? It's a quick change, at least.

README_Tradeoffs.md
michaeltlombardi: Which are, as far as I know, only delivered via PowerShell modules. This could use clarification though, something like

`...installing PowerShell modules containing the DSC Resources you wish to use...`

jpogran: When we refer to a single 'egg', do we call it 'carton'? Is a 'page' referred to as a 'book'? The delivery method or the container simply describe the method of transport or the type of conveyance, not the actual object.

DSC Resources use the PowerShell Module structure as a packaging method, but do not function like a PowerShell Module. Referring to both with one term that only applies to a single one, conflates the two when there is no reason to.

If anything, when I read that sentence I am told that I have to install 'PowerShell Modules' to use dsc_lite, which misses the very large point that DSC Resources are not bundled into the dsc_lite module nor are they 'downloaded' by the dsc_lite module

README_Tradeoffs.md
michaeltlombardi: I should clarify, the original authoring of this document assumed the audience was for users, especially moderate-advanced, of the existing implementation - the people most likely to experiment with a pre-release alternative.

I'll definitely go through and flesh out these factors (pluginsync, vendoring, parse-time safety).

michaeltlombardi:
Working on updates per @jpogran's feedback now.

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+The latter is a generalized resource that allows you to use _any_ DSC Resource available on the machine that you are configuring, while the former only allows you to use the DSC Resources bundled with the Puppet `dsc` module.
+
+## Why should you use the new `dsc_lite` module?
+
+Use `dsc_lite` if one or more of the following apply to you:
+
++ You need to use one or more DSC Resources which are _not_ [vendored](#vendoring) into Puppet's `dsc` module
++ You want to use class-based or composite DSC Resources
```
Iristyle: We might want to omit `composite` for now? I don't think anyone has tested this, and I thought @jpogran was a bit uncertain if we could do it? (I believe `Invoke-DscResource` can't be used on composites)

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
++ You need to use multiple versions of the same DSC resource
++ You want pluginsync to benefit from a [reduced footprint](#pluginsync-impact) - the file count for the module is reduced from thousands of files to a handful
++ You don't want to build your own custom version of the Puppet `dsc` module to include custom resources
++ You can live without the [validation](#validation) of the existing Puppet `dsc` module
++ You don't mind managing the installation of PowerShell modules containing DSC Resources on your nodes
++ You want to use the latest versions of DSC Resources even they're not [vendored](#vendoring)
```
Iristyle: Grammar

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Why should you keep using the Puppet `dsc` module?
+
+You should consider continuing to use the existing Puppet `dsc` module if one or more of the following apply to you:
+
++ You want the [validation](#validation) of the Puppet `dsc` Module
```
Iristyle: I might phrase this like "earlier feedback about the validity of manifest code"

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
+ Buffer:
+irectResourceAccess";
+};^
+
+insta
+
+Notice: Applied catalog in 3.33 seconds
+```
+
+Because `dsc_lite` is unaware of the types and requirements of any given DSC Resource, the module can't provide validation without actualling calling the code.
```
Iristyle: `actualling`

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
+  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_properties => {
+    ensure          => 'present',
+    content         => 'Apple, Banana, Cherry.',
+    destinationpath => 'C:\\Fruit.txt',
+  }
+}
+```
+
+The latter is a generalized resource that allows you to use _any_ DSC Resource available on the machine that you are configuring, while the former only allows you to use the DSC Resources bundled with the Puppet `dsc` module.
```
Iristyle: We might have to `*` the `any` later, but probably OK for now. As an example, I'm not sure if it makes sense to allow using `WaitForAny`, `WaitForAll`, `WaitForSome`(which we intentionally exclude from the `dsc` module) since we're generally not interested in building DSC dependency graphs.

michaeltlombardi: I think we should address this in the actual README, but not in the tradeoffs - have people tried to specify any of those when using Puppet?

I honestly didn't even think about them though, so good catch!

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
++ You can live without the [validation](#validation) of the existing Puppet `dsc` module
++ You don't mind managing the installation of PowerShell modules containing DSC Resources on your nodes
++ You want to use the latest versions of DSC Resources even if they're not [vendored](#vendoring)
+
+## Why should you keep using the Puppet `dsc` module?
+
+You should consider continuing to use the existing Puppet `dsc` module if one or more of the following apply to you:
+
++ You want [earlier feedback about the validity of manifest code](#validation)
++ You're only using Microsoft-provided DSC Resources - ie, not using custom, community-maintained, or class-based DSC Resources
```
clairecadman: i.e.*

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
+
+The existing Puppet `dsc` module vendors numerous DSC Resources into it - that is, the Puppet `dsc` module includes [many PowerShell DSC Resources](https://github.com/puppetlabs/puppetlabs-dsc/tree/master/lib/puppet_x/dsc_resources) inside it.
+This means that those DSC Resources are automatically on every machine which has the Puppet `dsc` module without you having to manage them individually.
+However, this means that if there are DSC Resources you do _not_ want to sync, or DSC Resources not vendored in the Puppet `dsc` module, you'll need to [rebuild the module to your own specifications](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/README_BUILD.md).
+
+The `dsc_lite` module does not vendor the DSC Resources for you, but neither does it limit you to only using vendored DSC Resources.
+Any DSC Resource available on the machine can be used by the `dsc_lite` module.
+
+#### Pluginsync impact
+Because the Puppet `dsc` module vendors many different DSC Resources it also has a large footprint and includes thousands of files.
```
clairecadman: Suggested reword: 

The Puppet `dsc` module vendors many different DSC Resources, and therefore has a large footprint and includes thousands of files.

README_Tradeoffs.md
```diff
@@ -0,0 +1,134 @@
...
+Because pluginsync downloads files from _every_ module in the agent's environment modulepath, regardless of whether the node uses them, this means that every machine using an environment that includes the Puppet `dsc` module will get these thousands of files synced to it.
```
clairecadman: Suggested reword:

The pluginsync downloads files from _every_ module in the agent's environment modulepath, regardless of whether the node uses them. This means that every machine using an environment that includes the Puppet `dsc` module will get these thousands of files synced to it.

README_Tradeoffs.md
glennsarti: I thought the preference from docs was not use abbreviations like i.e. or e.g..  Use the full "That is" or "For example"

clairecadman: Yes you are right actually, spelled out is preferred. @michaeltlombardi see comment above.

PR #19, opened 2018-01-09
(MODULES-6378) Update readme for dsc_lite changes

jpogran:
This commit updates the readme to reflect that it is for the dsc_lite
module and not the dsc module it was taken from. It removes all dsc module
related content and updates it for content related to dsc_lite.

Note: This must be merged after PR #16 and #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,33 +25,23 @@
 
 ## Description
 
-The Puppet dsc module manages Windows PowerShell DSC (Desired State Configuration) resources.
-
-This module generates Puppet types based on DSC Resources MOF (Managed Object Format) schema files.
-
-In this version, the following DSC Resources are already built and ready for use:
-
-- All base DSC resources found in PowerShell 5 ([WMF 5.0][wmf-5.0]).
-- All DSC resources found in the [Microsoft PowerShell DSC Resource Kit][DSCResources]
+The Puppet dsc_lite module allows you to manage target nodes using Windows PowerShell DSC (Desired State Configuration) Resources.
 
 ## Windows System Prerequisites
 
- - PowerShell 5, which is included in [Windows Management Framework 5.0][wmf-5.0].
- - [Windows 2003 is not supported](#known-issues).
+ - At least PowerShell 5.0, which is included in [Windows Management Framework 5.0][wmf-5.0].
 
 ## Setup
 
 ~~~bash
-puppet module install puppetlabs-dsc
+puppet module install puppetlabs-dsc_lite
 ~~~
 
 See [known issues](#known-issues) for troubleshooting setup.
 
 ## Usage
 
-### Using PSCredential or MSFT_Credential
-
-The generic `dsc` type is a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. You can use a DSC Resource by supplying the same properties you would set in a DSC Configuration script inside the `dsc_resource_properties` parameter. For most use cases the `dsc_resource_properties` parameter accepts the same structure as the PowerShell syntax, with the substituion of Puppet syntax for arrays, hashes and other data structures.
+The generic `dsc` type is a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. You can use a DSC Resource by supplying the same properties you would set in a DSC Configuration script inside the `dsc_resource_properties` parameter. For most use cases the `dsc_resource_properties` parameter accepts the same structure as the PowerShell syntax, with the substitution of Puppet syntax for arrays, hashes and other data structures.
 
 So a DSC resource specified in PowerShell...
 
@@ -82,7 +72,7 @@ The `dsc_resource_properties` parameter will recognize any key with a hash value
 A contrived, but simple example follows:
 
 ~~~puppet
-dsc{'foo':
+dsc {'foo':
   dsc_resource_name        => 'xFoo',
   dsc_resource_module_name => 'xFooBar',
   dsc_resource_properties  => {
@@ -103,7 +93,7 @@ dsc{'foo':
 Specifying credentials in DSC Resources requires using a PSCredential object. The `dsc` type will automatically create a PSCredential if the `dsc_type` has `MSFT_Credential` as a value.
 
 ~~~puppet
-dsc{'foouser':
+dsc {'foouser':
   dsc_resource_name        => 'User',
   dsc_resource_module_name => 'PSDesiredStateConfiguration',
   dsc_resource_properties  => {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -409,16 +319,17 @@ You can learn more about PowerShell DSC from the following online resources:
...
 - [Puppet Inc Windows DSC & WSUS Webinar 9-17-2015 webinar](https://puppet.com/webinars/windows-dsc-wsus-webinar-09-17-2015) - How DSC works with Puppet
 - [Better Together: Managing Windows with Puppet, PowerShell and DSC - PuppetConf 10-2015 talk](https://www.youtube.com/watch?v=TP0zqe-yQto) and [slides](https://speakerdeck.com/iristyle/better-together-managing-windows-with-puppet-powershell-and-dsc)
 - [PowerShell.org](http://powershell.org/wp/tag/dsc/) - Community based DSC tagged posts
 - [PowerShell Magazine](http://www.powershellmagazine.com/tag/dsc/) - Community based DSC tagged posts
 
 There are several books available as well. Here are some selected books for reference:
 
-- [Learning PowerShell DSC](http://bit.ly/learndsc) - James Pogran is a member of the team here at Puppet Inc working on the DSC/Puppet integration
+- [Learning PowerShell DSC 2nd Edition](https://www.packtpub.com/networking-and-servers/learning-powershell-dsc-second-edition) - James Pogran is a member of the team here at Puppet Inc working on the DSC/Puppet integration
```
Iristyle: Get that 💰 !

README.md
```diff
@@ -25,32 +25,22 @@
...
 
-The Puppet dsc module manages Windows PowerShell DSC (Desired State Configuration) resources.
-
-This module generates Puppet types based on DSC Resources MOF (Managed Object Format) schema files.
-
-In this version, the following DSC Resources are already built and ready for use:
-
-- All base DSC resources found in PowerShell 5 ([WMF 5.0][wmf-5.0]).
-- All DSC resources found in the [Microsoft PowerShell DSC Resource Kit][DSCResources]
+The Puppet dsc_lite module allows you to manage targe nodes using Windows PowerShell DSC (Desired State Configuration) Resources.
```
Iristyle: `targe` -> `target`

README.md
```diff
@@ -237,83 +229,13 @@ To express EmbeddedInstances, the `dsc_resource_properties` parameter will recog
   ~~~
 
 - You cannot use forward slashes for the MSI `Path` property for the `Package` DSC Resource. The underlying implementation does not accept forward slashes instead of backward slashes in paths, and it throws a misleading error that it could not find a Package with the Name and ProductId provided. [MODULES-2486](https://tickets.puppet.com/browse/MODULES-2486) has more examples and information on this subject.
-- `dsc_ensure` overrides and ignores the value in `ensure` if both are present in a Puppet DSC resource. See [Using DSC Resources with Puppet](#using-dsc-resources-with-puppet).
```
Iristyle: This is one we need to file a ticket in dsc_lite for BTW ... we have a similar problem around rationalizing Puppet `ensure` vs the DSC resource `ensure`.  The Beaker tests document some of this behavior - we'll probably want some new tests documenting expected behaviors.

README.md
```diff
@@ -237,83 +229,13 @@ To express EmbeddedInstances, the `dsc_resource_properties` parameter will recog
...
+
 - Use of this module with the 3.8.x x86 version of Puppet is highly discouraged, though supported.  Normally, this module employs a technique to dramatically improve performance by reusing a PowerShell process to execute DSC related commands.  However, due to the Ruby 1.9.3 runtime used with the 3.8.x x86 version of Puppet, this technique must be disabled, resulting in at least a 2x slowdown.
 
 ### Known Issues
 
-- The `dsc_log` resource might not appear to work. The ["Log" resource](https://technet.microsoft.com/en-us/library/Dn282117.aspx) writes events to the 'Microsoft-Windows-Desired State Configuration/Analytic' event log, which is [disabled by default](https://technet.microsoft.com/en-us/library/Cc749492.aspx).
```
Iristyle: I suspect (most of?) these resource specific problems are all still issues with `dsc`, but we can revisit / decide if we want to bring any of these examples back later I think. I know you're not keen on documenting DSC resource bugs... but I think our thinking was to curb support tix.

README.md
jpogran: I did update the link for the other book...

README.md
jpogran: The text as is is incorrect, so I removed it. I think figuring out the complex interaction between `ensure` in the Puppet provider and `Ensure` in the DSC Resource, nevermind the case where a user might not sync the two, is outside the scope of this unsupported release

README.md
jpogran: They all refer to types in the dsc module, not the dsc_lite module, and refer to problems with DSC and not Puppet. We warned that there aren't guardrails with this module, so I don't see the need to duplicate the list from the dsc module

jpogran:
All typos addressed, and pulled in suggestions from #16

README.md
Iristyle: Yup, that's why I was saying we should file a ticket.

README.md
Iristyle: The two modules have different READMEs / repos / forge landing pages, so if we care about a problem with `dsc_log {}` and `dsc { dsc_resource_name => 'log' }` has the same problem, I think it would help users out.  But if we don't care about it / think its helpful, should probably remove from the other DSC module README as well.

README.md
```diff
@@ -125,12 +115,12 @@ dsc{'foouser':
 
 ### Using CimInstance
 
-A DSC Resource may need a more complex type that a simple key value pair, so an EmbeddedInstance is used. An EmbeddedInstance is serialized as CimInstance over the wire. In order to represent a CimInstance in the `dsc` type, we will use the `dsc_type` key to specify which CimInstance to use. If the CimInstance is an array, we append a `[]` to the end of the name.
+A DSC Resource may need a more complex type than a simple key value pair, so an EmbeddedInstance is used. An EmbeddedInstance is serialized as CimInstance over the wire. In order to represent a CimInstance in the `dsc` type, we will use the `dsc_type` key to specify which CimInstance to use. If the CimInstance is an array, we append a `[]` to the end of the name.
```
Iristyle: Still could use a little background on EmbeddedInstance / CimInstance here I think, though I couldn't find great MSDN docs on either.  

Only found CDXML page at https://msdn.microsoft.com/en-us/library/jj542521%28v=vs.85%29.aspx which isn't awesome.

PR #21, opened 2018-01-10
(MAINT) Clarify tradeoffs wording

michaeltlombardi:
Some of the wording the tradeoffs is unclear or
grammatically incorrect. This maintenance commit
fixes those problems.

CC @clairecadman

### DIFF
--- a/README_Tradeoffs.md
+++ b/README_Tradeoffs.md
@@ -53,7 +53,7 @@ Use `dsc_lite` if one or more of the following apply to you:
 You should consider continuing to use the existing Puppet `dsc` module if one or more of the following apply to you:
 
 + You want [earlier feedback about the validity of manifest code](#validation)
-+ You're only using Microsoft-provided DSC Resources - ie, not using custom, community-maintained, or class-based DSC Resources
++ You're only using Microsoft-provided DSC Resources - i.e., not using custom, community-maintained, or class-based DSC Resources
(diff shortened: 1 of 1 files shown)


README_Tradeoffs.md
```diff
@@ -77,8 +77,9 @@ The `dsc_lite` module does not vendor the DSC Resources for you, but neither doe
 Any DSC Resource available on the machine can be used by the `dsc_lite` module.
 
 #### Pluginsync impact
-Because the Puppet `dsc` module vendors many different DSC Resources it also has a large footprint and includes thousands of files.
-Because pluginsync downloads files from _every_ module in the agent's environment modulepath, regardless of whether the node uses them, this means that every machine using an environment that includes the Puppet `dsc` module will get these thousands of files synced to it.
+The Puppet `dsc` module vendors many different DSC Resources, and therefore has a large footprint and includes thousands of files.
+The pluginsync downloads files from every module in the agent's environment modulepath, regardless of whether the node uses them.
+This means that _every_ machine using an environment that includes the Puppet `dsc` module will get these thousands of files synced to them.
```
clairecadman: *...module will get thousands of these files...etc?

michaeltlombardi: Hmmm, unsure here - we're referencing specifically the thousands of files that come with the Puppet `dsc` module. 

I'm not attached to the wording, but does `thousands of these files` and `these thousands of files` have a semantic difference? In my head they _seem_ to.

clairecadman: hmm yes I think they do too. Does "the thousands of files" work?

michaeltlombardi:
@clairecadman, updated the wording one last time. :)

PR #28, opened 2018-01-30
(maint) Beaker test suite cleanup

Iristyle:
This is a redo of #26, now with passing code. That PR was errantly merged prior to verification, and triggered CI failures due to test bugs introduced. #27 reverts that code so that we can start fresh with this PR, which *has* been run locally first.


ThoughtCrhyme:
Yay for locally passing runs!

PR #29, opened 2018-01-31
(MODULES-5845) Add DSC Resource ModuleSpecification support

jpogran:
This commit changes `dsc_resource_module_name` to `dsc_resource_module`, and updates it to accept both a string or a hash with two specific key/value pairs

- [x] Readme
- [x] Unit tests
- [x] Update acceptance tests based on parameter changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,13 @@
 
 #### Table of Contents
 
-1. [Description - What is the dsc module and what does it do](#module-description)
+1. [Description - What is the dsc module and what does it do](#description)
 2. [Prerequisites](#windows-system-prerequisites)
 3. [Setup](#setup)
 4. [Usage](#usage)
-  * [Using PSCredential or MSFT_Credential](#using-dsc-resources-with-puppet)
-  * [Using EmbeddedInstance or CimInstance](#using-dsc-resources-with-puppet)
+  * [Specifying a DSC Resource version](#specifying-a-dsc-resource-version)
+  * [Using PSCredential or MSFT_Credential](#using-pscredential-or-msft_credential)
+  * [Using EmbeddedInstance or CimInstance](#using-ciminstance)
   * [Handling Reboots with DSC](#handling-reboots-with-dsc)
 5. [Reference](#reference)
   * [Types and Providers](#types-and-providers)
@@ -20,7 +21,7 @@
   * [Known Issues](#known-issues)
   * [Running Puppet and DSC without Administrative Privileges](#running-puppet-and-dsc-without-administrative-privileges)
 7. [Development - Guide for contributing to the module](#development)
-8. [Places to Learn More About DSC](#places-to-learn-more-about-dsc)
+8. [Learn More About DSC](#learn-more-about-dsc)
 9. [License](#license)
 
 ## Description
@@ -57,7 +58,7 @@ WindowsFeature IIS {
 ~~~puppet
 dsc {'iis':
   dsc_resource_name        => 'WindowsFeature',
-  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_module => 'PSDesiredStateConfiguration',
   dsc_resource_properties  => {
     ensure => 'present',
     name   => 'Web-Server',
@@ -74,7 +75,7 @@ A contrived, but simple example follows:
 ~~~puppet
 dsc {'foo':
   dsc_resource_name        => 'xFoo',
-  dsc_resource_module_name => 'xFooBar',
+  dsc_resource_module => 'xFooBar',
   dsc_resource_properties  => {
     ensure  => 'present',
     fooinfo => {
@@ -88,6 +89,24 @@ dsc {'foo':
 }
 ~~~
 
+### Specifying a DSC Resource version
+
+When there is more than one version installed for a given DSC Resource module, you must specify the version in your declaration. You can specify the version with similar syntax to a DSC Configuration script by using a hash containing the name and version of the DSC Resource module to use.
+
+~~~puppet
+dsc {'iis_server':
+  dsc_resource_name   => 'WindowsFeature',
+  dsc_resource_module => {
+    name    => 'PSDesiredStateConfiguration',
+    version => '1.1'
+  },
+  dsc_resource_properties  => {
+    ensure => 'present',
+    name   => 'Web-Server',
+  }
+}
+~~~
+
 ### Using PSCredential or MSFT_Credential
 
 Specifying credentials in DSC Resources requires using a PSCredential object. The `dsc` type will automatically create a PSCredential if the `dsc_type` has `MSFT_Credential` as a value.
@@ -95,7 +114,7 @@ Specifying credentials in DSC Resources requires using a PSCredential object. Th
 ~~~puppet
 dsc {'foouser':
   dsc_resource_name        => 'User',
-  dsc_resource_module_name => 'PSDesiredStateConfiguration',
+  dsc_resource_module => 'PSDesiredStateConfiguration',
   dsc_resource_properties  => {
     'username'    => 'jane-doe',
     'description' => 'Jane Doe user',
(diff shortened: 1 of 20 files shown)


tests/acceptance/tests/dsc_type/multiple_dsc_resources_in_PSModulePath.rb
```diff
@@ -108,20 +110,12 @@
...
 
     step 'Verify Results'
-    # PuppetFakeResource always overwrites file at this path
-    # PuppetFakeResourc2 2.0 appends "v2" to the written file before "ImportantStuff"
-    on(agent, "cat /cygdrive/c/#{fake_name}", :acceptable_exit_codes => [1]) do |result|
-      expect_failure('Cannot yet execute PuppetFakeResource 2.0 until MODULES-5845 implemented') do
-        assert_match(/^v2#{test_file_contents}/, result.stdout, 'PuppetFakeResource File contents incorrect!')
-      end
-    end
+    # if this file exists, 'absent' didn't work
```
Iristyle: I think you want to actually make sure the file exists here and that it contains the right data.  The assertion works because the v2 type appends the string "v2" to the front of the output.


```ruby
on(agent, "cat /cygdrive/c/#{fake_name}", :acceptable_exit_codes => [0]) do |result|
  assert_match(/^v2#{test_file_contents}/, result.stdout, 'PuppetFakeResource File contents incorrect!')
end
```

README.md
```diff
@@ -11,8 +11,9 @@
 2. [Prerequisites](#windows-system-prerequisites)
 3. [Setup](#setup)
 4. [Usage](#usage)
-  * [Using PSCredential or MSFT_Credential](#using-dsc-resources-with-puppet)
-  * [Using EmbeddedInstance or CimInstance](#using-dsc-resources-with-puppet)
+  * [Specifying a DSC Resource version](#specifying-a-dsc-resource-version)
+  * [Using PSCredential or MSFT_Credential](#using-pscredential-or-msft_credential)
+  * [Using EmbeddedInstance or CimInstance](#using-embeddedinstance-or-ciminstance)
```
Iristyle: good catch on the anchors

README.md
```diff
@@ -88,14 +89,48 @@ dsc {'foo':
...
+}
+~~~
+
+Specifying the path to the DSC Resource module works as well:
+
+~~~puppet
+dsc {'iis_server':
+  dsc_resource_name   => 'WindowsFeature',
+  dsc_resource_module => {
+    name    => 'C:\Windows\system32\WindowsPowershell\v1.0\Modules\PsDesiredStateConfiguration\PSDesiredStateConfiguration.psd1',
```
Iristyle: Will test locally as well, but is it the path to the `PSD1` that's used, or the path to the directory?

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -69,27 +69,39 @@
...
+
+    it "should require name and version keys if hash" do
+      expect {
+        resource[:dsc_resource_module] = { 'foo' => 'bar'}
+      }.to raise_error(Puppet::Error, /Must specify name and version if using ModuleSpecification/)
+    end
+
     it "should not allow nil" do
       expect {
-        resource[:name] = nil
```
Iristyle: Whooooops

README.md
Iristyle: So it looks like DSC allows you to use a parent directory instead of the PSD1, but then that allows it to find the wrong thing.

I created a gist demonstrating all the various scenarios for side-by-side versions of the same module at:
https://gist.github.com/Iristyle/92a722ee0bfe61aa7ee5ec53dbf65b4a

Fully qualified path + version kind of doesn't make sense to me, given the above gist shows that the path already has to point to the specific version when side-by-side DSC modules are in play. And when the parent directory of all the side-by-side modules was specified, it used the wrong version *and* didn't generate an error!

If the example above aligns with the system I'm testing, the PSD1 has `1.1` in it, which makes it simpler for an end user to just specify `dsc_resource_module` as a string, unless they really need it there for documentation purposes?

```puppet
dsc_resource_module => 'C:\Windows\system32\WindowsPowershell\v1.0\Modules\PsDesiredStateConfiguration\PSDesiredStateConfiguration.psd1'
```

Basically I'm saying I wouldn't recommend users do this given it's more to put in a manifest than is necessary and has the potential to deviate from expectations if a user puts directory instead of PSD1 file.

Iristyle: Interestingly, using your example with `PSDesiredStateConfiguration` the behavior is a little different. It won't load an invalid version, but not sure if that's because it's not one of the side-by-side style installations like I used above?

Trying to load an invalid version 1.5 using the parent directory fails as you would expect:

```
PS C:\Users\Administrator> Invoke-DscResource -ModuleName @{ ModuleName = "C:\Windows\System32\WindowsPowerShell\v1.0\Modules\PSDesiredStateConfiguration"; ModuleVersion = '1.5' } -Name WindowsFeature -Property @{ Ensure = 'present'; Name = 'Web-Server' } -Method Set
Invoke-DscResource : Resource WindowsFeature was not found.
At line:1 char:1
+ Invoke-DscResource -ModuleName @{ ModuleName = "C:\Windows\System32\W ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Invoke-DscResource], ArgumentException
    + FullyQualifiedErrorId : System.ArgumentException,Microsoft.PowerShell.DesiredStateConfiguration.Commands.InvokeDscResourceMethodCommand
```

And using the correct 1.1 succeeds in executing as you would expect

```
PS C:\Users\Administrator> Invoke-DscResource -ModuleName @{ ModuleName = "C:\Windows\System32\WindowsPowerShell\v1.0\Modules\PSDesiredStateConfiguration"; ModuleVersion = '1.1' } -Name WindowsFeature -Property @{ Ensure = 'present'; Name = 'Web-Server' } -Method Set
WARNING: [E3HEJGHNN47UI95]:                            [[WindowsFeature]DirectResourceAccess] Failed to start automatic updating for installed components. Error: 0x80070422

RebootRequired
--------------
False
```

Note I'm not fully qualifying the PSD1

jpogran: As with how we do it in the dsc module, it has to be the fully qualified path to the psd1 with the psd1 specified, not just the module directory for it to work.

It was a mistake to document using the version as well as the path, as specifying the path is only really useful if the DSC Resource is not in the default paths because you can't rely on the engine to find it. The version is immaterial at this point.

Instead of investigating the vagaries of how Invoke-DscResource, I think it better we stick to documenting what the source docs show (name and version only) and leave it at that.

Iristyle: Yup, +1 ... that's basically what I was saying above. This is a weird / confusing usage, and we're probably better off ignoring it.

README.md
Iristyle: Looks like further up `description` has the same problem - we can fix in a follow-up.

tests/acceptance/tests/dsc_type/multiple_dsc_resources_in_PSModulePath.rb
```diff
@@ -108,20 +110,14 @@
...
     end
 
     step 'Verify Results'
-    # PuppetFakeResource always overwrites file at this path
-    # PuppetFakeResourc2 2.0 appends "v2" to the written file before "ImportantStuff"
-    on(agent, "cat /cygdrive/c/#{fake_name}", :acceptable_exit_codes => [1]) do |result|
-      expect_failure('Cannot yet execute PuppetFakeResource 2.0 until MODULES-5845 implemented') do
-        assert_match(/^v2#{test_file_contents}/, result.stdout, 'PuppetFakeResource File contents incorrect!')
-      end
+    # if this file exists, 'absent' didn't work
```
Iristyle: Looks like you didn't carry over the comment from the prior version / in my fixup commit https://github.com/jpogran/puppetlabs-dsc_lite/pull/1/commits/fdc8d834c212862bdb47025f6505d109963986d0#diff-8ed1e98461c1f6d1b58dcb66de5b2427R118

PR #30, opened 2018-01-31
(MODULES-6548) Move Beaker integrations to acceptance

Iristyle:
This PR is another in the line of moving towards beaker-rspec and testmode_switcher.

"Integration" tests are master / agent based tests and not currently executing in CI. "Acceptance" tests are agent only using `puppet apply`.

Most master specific code has been removed, though in some cases some notes or code modifications were made for future support of testmode_switcher, most notably with respect to pathing when `PuppetFakeResource` is copied to SUTs.

PR #32, opened 2018-02-01
(MODULES-6548) Re-add empty file to "integration"

Iristyle:
- Current CI configuration requires files be present in the
   tests/integration/tests folder. To appease Father Jenkins, add a
   nominal file to the directory until such time that the pipelines may
   be reconfigured.


Iristyle:
Presumably required to address the CI failure introduced in #30

PR #36, opened 2018-02-07
(maint) Allow reboot 2.0+ module dependency

TraGicCode:
I had been using the 2.0.0 version of the reboot module and noticed that when using PMT to install dsc_lite it blows up because the module i have locally doesn't fall within an acceptable version for dsc_lite as specified in the metadata.json.  Just bumping the dependency and would like someone to review to make sure they don't forsee any issues with this.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -35,7 +35,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/reboot",
-      "version_requirement": ">= 1.2.1 < 2.0.0"
+      "version_requirement": ">= 1.2.1 < 3.0.0"
     }
   ]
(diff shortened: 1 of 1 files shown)


Iristyle:
Thanks for the PR @TraGicCode ... this should be fine. Would you mind updating commit subject to follow our standard for maintenance commits (start with `(maint)`, max length of 50)?  Something like:

```
(maint) Allow reboot 2.0 module dependency
```

Thanks!

TraGicCode:
@Iristyle Done.

Iristyle:
Thanks @TraGicCode !

PR #37, opened 2018-02-21
(WIP) (MODULES-6450) Add Beaker test for installing DSC resources via module

RandomNoun7:
This PR adds helpers for packaging the included PuppetFakeResource as a module
that can be installed into a versioned directory via the Install-Module Command.
It then impelements a test to ensure that the module can be successfully
imported and used in a manifest.

### DIFF
--- a/tests/acceptance/tests/basic_functionality/puppet_apply_package_installed_module.rb
+++ b/tests/acceptance/tests/basic_functionality/puppet_apply_package_installed_module.rb
@@ -1,0 +1,63 @@
+require 'erb'
+require 'dsc_utils'
+require 'securerandom'
+test_name 'MODULES-6450 - Add beaker tests for installing DSC Resources via Install-Module cmdlet'
+
+installed_path = get_fake_reboot_resource_install_path(usage = :manifest)
+
+confine(:to, :platform => 'windows')
+
+test_dir_path      = SecureRandom.uuid
+fake_name          = SecureRandom.uuid
+test_file_contents = SecureRandom.uuid
+repo_name          = SecureRandom.uuid
+repo_folder        = SecureRandom.uuid
+
+dsc_manifest = <<-MANIFEST
+file { 'C:/#{ test_dir_path }' :
+   ensure => 'directory'
+}
+->
+dsc { '#{fake_name}':
+  dsc_resource_name => 'puppetfakeresource',
+  dsc_resource_module => '#{installed_path}/PuppetFakeResource',
+  dsc_resource_properties => {
+    ensure          => 'present',
+    importantstuff  => '#{test_file_contents}',
+    destinationpath => '#{"C:\\" + test_dir_path + "\\" + fake_name}',
+  },
+}
+MANIFEST
+
+# Teardown
+teardown do
+  step 'Remove Test Artifacts'
+  on(agents, "rm -rf /cygdrive/c/#{test_dir_path}")
+  agents.each do |agent|
+    uninstall_fake_reboot_resource(agent)
+  end
+end
+
+# Tests
+agents.each do |agent|
+  step 'Create local Nuget repository'
+  create_local_nuget_repo(agent, repo_name, repo_folder)
+
+  step 'Publish PuppetFakeResource module to the repo'
+  publish_dsc_module_to_nuget(agent, repo_name)
+  
+  step 'Install the module via Install-Module cmdlet'
+  powershell_install_dsc_module(agent, 'puppetfakeresource', repo_name)
+
+  step 'Apply Manifest'
+  on(agent, puppet('apply'), :stdin => dsc_manifest, :acceptable_exit_codes => [0,2]) do |result|
+    assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    assert_match(/Stage\[main\]\/Main\/Dsc\[#{fake_name}\]\/ensure\: created/, result.stdout, 'DSC Resource missing!')
+  end
+
+  step 'Verify Results'
+  # PuppetFakeResource always overwrites file at this path
+  on(agent, "cat /cygdrive/c/#{test_dir_path}/#{fake_name}", :acceptable_exit_codes => [0]) do |result|
+    assert_match(/#{test_file_contents}/, result.stdout, 'PuppetFakeResource File contents incorrect!')
+  end
+end
--- a/tests/lib/dsc_utils.rb
+++ b/tests/lib/dsc_utils.rb
@@ -92,6 +92,45 @@ def install_fake_reboot_resource(host)
   end
 end
 
+def create_local_nuget_repo(host, repo_name, repo_folder)
+  create_repo_command = "Register-PSRepository "
+  create_repo_command << "-Name #{repo_name} "
+  create_repo_command << "-SourceLocation c:\\\\temp\\\\#{repo_folder} "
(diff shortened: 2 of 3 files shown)


tests/lib/dsc_utils.rb
```diff
@@ -92,6 +92,45 @@ def install_fake_reboot_resource(host)
...
+  create_repo_command << "-Name #{repo_name} "
+  create_repo_command << "-SourceLocation c:\\\\temp\\\\#{repo_folder} "
+  create_repo_command << "-ScriptSourceLocation c:\\\\temp\\\\#{repo_folder} "
+  create_repo_command << "-InstallationPolicy Trusted"
+
+  bootstrap_nuget = 'Install-PackageProvider -Name NuGet -MinimumVersion "2.8.5.201" -Force'
+
+  on(host, powershell("New-Item c:\\\\temp\\\\#{repo_folder} -ItemType Directory"))
+  on(host,powershell(bootstrap_nuget))
+  on(host, powershell(create_repo_command))
```
Iristyle: Does having this in 3 separate command invocations make debugging easier?

tests/lib/dsc_utils.rb
```diff
@@ -92,6 +92,45 @@ def install_fake_reboot_resource(host)
...
+end
+
+# This command will automatically install nuget on the host SilentlyContinue is
+# there to suppress warnings about certain properties not being present in the
+# module psd1
+def publish_dsc_module_to_nuget(host, repo_name)
+  fake_reboot_resource_source_path = "tests/files/dsc_puppetfakeresource/PuppetFakeResource"
+
+  scp_to(host, fake_reboot_resource_source_path, 'c:/temp')
```
Iristyle: Should probably randomize `c:/temp`

tests/lib/dsc_utils.rb
```diff
@@ -92,6 +92,45 @@ def install_fake_reboot_resource(host)
...
+
+  publish_command = "$env:TEMP = 'c:\\temp'; "
+  publish_command << "Publish-Module "
+  publish_command << "-Path c:/temp/PuppetFakeResource "
+  publish_command << "-Repository #{repo_name} "
+  publish_command << "-Force "
+  publish_command << "-WarningAction SilentlyContinue"
+
+  on(host, powershell(publish_command, {'EncodedCommand' => true}))
```
Iristyle: If `c:/temp` is used only as a holding ground prior to actual publishing, I think it can be trashed immediately after module is published

tests/lib/dsc_utils.rb
```diff
@@ -92,6 +92,45 @@ def install_fake_reboot_resource(host)
...
+end
+
+def powershell_install_dsc_module(host, module_name, repo_name)
```
Iristyle: I think this should optionally take the version flag as well.

tests/lib/dsc_utils.rb
```diff
@@ -92,6 +92,45 @@ def install_fake_reboot_resource(host)
...
 end
 
+def create_local_nuget_repo(host, repo_name, repo_folder)
+  create_repo_command = "Register-PSRepository "
+  create_repo_command << "-Name #{repo_name} "
+  create_repo_command << "-SourceLocation c:\\\\temp\\\\#{repo_folder} "
+  create_repo_command << "-ScriptSourceLocation c:\\\\temp\\\\#{repo_folder} "
+  create_repo_command << "-InstallationPolicy Trusted"
+
+  bootstrap_nuget = 'Install-PackageProvider -Name NuGet -MinimumVersion "2.8.5.201" -Force'
```
Iristyle: Is there any way we can prevent this network dependency (in a subsequent iteration)? 

If we can put the module zip inside the repo, I think that would help.

tests/acceptance/tests/basic_functionality/puppet_apply_package_installed_module.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+end
+
+# Tests
+agents.each do |agent|
+  step 'Create local Nuget repository'
+  create_local_nuget_repo(agent, repo_name, repo_folder)
+
+  step 'Publish PuppetFakeResource module to the repo'
+  publish_dsc_module_to_nuget(agent, repo_name)
```
Iristyle: Probably want a version here

tests/acceptance/tests/basic_functionality/puppet_apply_package_installed_module.rb
```diff
@@ -0,0 +1,63 @@
...
+  
+  step 'Install the module via Install-Module cmdlet'
+  powershell_install_dsc_module(agent, 'puppetfakeresource', repo_name)
```
Iristyle: And a version here as well

tests/acceptance/tests/basic_functionality/puppet_apply_package_installed_module.rb
```diff
@@ -0,0 +1,63 @@
...
+}
+MANIFEST
+
+# Teardown
+teardown do
+  step 'Remove Test Artifacts'
+  on(agents, "rm -rf /cygdrive/c/#{test_dir_path}")
+  agents.each do |agent|
+    uninstall_fake_reboot_resource(agent)
+  end
```
Iristyle: Do we want to remove the module from our local Nuget feed after we're done using it?

tests/lib/dsc_utils.rb
```diff
@@ -92,6 +92,45 @@ def install_fake_reboot_resource(host)
   end
 end
 
+def create_local_nuget_repo(host, repo_name, repo_folder)
```
Iristyle: I don't think we need `repo_folder` as a parameter... can just reuse `repo_name`.

tests/lib/dsc_utils.rb
RandomNoun7: Slightly, keeps my functions smaller. Mostly though it's just the way I stubbed out the functions based the todo list notes I was keeping on how to accomplish this task. This code so far is a barely working proof to myself that the idea will work with little thought given to the way the code should actually look and flow.

tests/lib/dsc_utils.rb
RandomNoun7: I can change this to a randomized folder name at the top level of c:\

tests/lib/dsc_utils.rb
RandomNoun7: I can make that a part of the teardown routine in the overall tests file.

tests/acceptance/tests/basic_functionality/puppet_apply_package_installed_module.rb
RandomNoun7: Seems like a good idea.

tests/lib/dsc_utils.rb
RandomNoun7: When I switch to a top level randomized folder I make this change.

tests/lib/dsc_utils.rb
RandomNoun7: Do you mean zip up the package provider and ship it with the module?

tests/lib/dsc_utils.rb
RandomNoun7: We can do that. I was considering a set of functions that would even take the existing PuppetFakeResource and create another copy of it with an incremented version number and publish that. Not sure if it's necessary yet or not.

PR #41, opened 2018-03-16
Modules 6860 add dsc lite feature improve confine messages

Iristyle:
Similar to https://github.com/puppetlabs/puppetlabs-dsc/pull/350 this is a quick fix that might require some further thought given there are currently two version checks against PowerShell performed in two different ways at different points in the lifecycle.

This builds on #40, but if this is merged first, #40 can be closed.

 - Previously when a version of PowerShell older than 5.0.10586.117
   was installed, Puppet would return an error message like:

   Error: Could not find a suitable provider for dsc_lite

   Unfortunately this message did not yield any useful information,
   even with --debug and --verbose turned on, because of the way the
   existing confine check was performed with a confine :true

   To produce a more useful message that is meaningful to users,
   introduce a new feature that performs the same check as the
   confine, but that also will issue an error message (only once)
   that indicates the installed version / required version constraints
   for the module, changing the output to:.

   Error: The dsc_lite module requires PowerShell version 5.0.10586.117 - current version 5.0.10514.6\n   (file & line not available)
   Error: Could not find a suitable provider for dsc_lite

 - Note that DSC_MODULE_POWERSHELL_UPGRADE_MSG is never actually shown
   when the PowerShell version is < 5.0.10586.117 because the check
   occurs too late in the lifecycle of loading providers. The confine
   code has already evaluated, preventing the other messsages from
   being shown.


RandomNoun7:
Previous error messge
```
Error: Could not find a suitable provider for dsc
```
New error message
```
Error: The dsc_lite module requires PowerShell version 5.0.10586.117 - current version 2.0\n   (file & line not available)
Error: Could not find a suitable provider for dsc
```

lib/puppet/feature/dsc_lite.rb
```diff
@@ -0,0 +1,20 @@
+require 'puppet/util/feature'
+
+if Puppet.features.microsoft_windows?
+  required_version = Gem::Version.new('5.0.10586.117')
+  installed_version = Gem::Version.new(Facter.value(:powershell_version))
+
+  if (installed_version >= required_version)
+    Puppet.features.add(:dsc_lite)
+  else
+    Puppet.warn_once(
```
Iristyle: The definition of `warn_once` is at https://github.com/puppetlabs/puppet/blob/80e7948c69b6c2145a709c6e7fc95bd7bb772f26/lib/puppet/util/logging.rb#L167

PR #51, opened 2018-06-01
(MODULES-7143) add UTF-8 encoding to ERB templates.

ThoughtCrhyme:
This PR adds UTF-8 support for fields inside of custom dsc templates.

It does not support the template custom name itself being in UTF-8, but parameters passed to fields in the template can now be UTF-8.

Also added an integration test for this functionality.

### DIFF
--- a/tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
+++ b/tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
@@ -1,0 +1,92 @@
+require 'erb'
+require 'master_manipulator'
+require 'dsc_utils'
+test_name 'Apply generic DSC Manifest with UTF-8 file name to create a puppetfakeresource'
+
+installed_path = get_fake_reboot_resource_install_path(usage = :manifest)
+
+# different UTF-8 widths
+# 1-byte A
+# 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
+# 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+# 4-byte 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+MIXED_UTF8 = "A\u06FF\u16A0\u{2070E}" # Aۿᚠ𠜎
+
+# Manifest
+file_name = MIXED_UTF8
+file_path = SecureRandom.uuid
+test_file_contents = SecureRandom.uuid
+dsc_manifest = <<-MANIFEST
+dsc {'some_name':
+  dsc_resource_name => 'puppetfakeresource',
+  # NOTE: install_fake_reboot_resource installs on master, which pluginsyncs here
+  dsc_resource_module => '#{installed_path}/PuppetFakeResource',
+  dsc_resource_properties => {
+    ensure          => 'present',
+    importantstuff  => '#{test_file_contents}',
+    destinationpath => 'C:\\#{file_path}\\#{file_name}'
+  }
+}
+MANIFEST
+
+# Teardown
+teardown do
+  confine_block(:to, :platform => 'windows') do
+    step 'Remove Test Artifacts'
+    agents.each do |agent|
+      uninstall_fake_reboot_resource(agent)
+    end
+    on(agents, "rm -rf C:/#{file_path}")
+  end
+end
+
+# Tests
+confine_block(:to, :platform => 'windows') do
+  agents.each do |agent|
+    step 'setup'
+    on(agent, powershell("mkdir /#{file_path}"))
+
+    step 'Copy Test Type Wrappers'
+    install_fake_reboot_resource(agent)
+
+    step 'Run Puppet Apply'
+    on(agent, puppet('apply'), :stdin => dsc_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    powershell_cmd = "'Write-Host (Get-ChildItem C:\\#{file_path}  -Filter '#{file_name}' | Measure-Object ).Count;'"
+    on(agent, powershell(powershell_cmd)) do |result|
+      assert('1' == result.stdout.to_s.strip, 'File with correct UTF-8 characters was not present.')
+    end
+  end
+end
+
+# New manifest to remove value.
+dsc_remove_manifest = <<-MANIFEST
+dsc {'some_name':
+  dsc_resource_name => 'puppetfakeresource',
+  dsc_resource_module => '#{installed_path}/PuppetFakeResource',
+  dsc_resource_properties => {
+    ensure          => 'absent',
+    importantstuff  => '#{test_file_contents}',
+    destinationpath => 'C:\\#{file_path}\\#{file_name}'
+  }
+}
+MANIFEST
+
+confine_block(:to, :platform => 'windows') do
+  agents.each do |agent|
+    step 'Apply Manifest to Remove File'
+    on(agent, puppet('apply'), :stdin => dsc_remove_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    # if the file count is greater than 0, 'absent' didn't work
+    powershell_cmd = "'Write-Host (Get-ChildItem C:\\#{file_path} -Filter '#{file_name}' | Measure-Object ).Count;'"
+    on(agent, powershell(powershell_cmd)) do |result|
+      assert('0' == result.stdout.to_s.strip, 'File with UTF-8 character name was not removed')
+    end
+  end
+end
--- a/lib/puppet/provider/base_dsc_lite/powershell.rb
+++ b/lib/puppet/provider/base_dsc_lite/powershell.rb
@@ -190,7 +190,8 @@ EOT
     template_name = resource.type == :dsc ?
       '/invoke_generic_dsc_resource.ps1.erb' :
       '/invoke_dsc_resource.ps1.erb'
-    template = ERB.new(File.new(template_path + template_name).read, nil, '-')
+    file = File.new(template_path + template_name, :encoding => Encoding::UTF_8)
+    template = ERB.new(file.read, nil, '-')
     template.result(binding)
   end
 end


tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,78 @@
+require 'erb'
+require 'master_manipulator'
+require 'dsc_utils'
+test_name 'Apply generic DSC Manifest with UTF-8 file name to create a puppetfakeresource'
+
+installed_path = get_fake_reboot_resource_install_path(usage = :manifest)
+
+# Manifest
+fake_name = "\u4388\u542B\u3D3C\u7F4D\uF961\u4381\u53F4\u79C0\u3AB2\u8EDE" # 䎈含㴼罍率䎁叴秀㪲軞
```
Iristyle: I believe all of these characters are the same byte width... can we instead use characters whose width vary?

The standard string we use in most places is 

```ruby
# different UTF-8 widths
  # 1-byte A
  # 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
  # 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
  # 4-byte 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
  MIXED_UTF8 = "A\u06FF\u16A0\u{2070E}" # Aۿᚠ𠜎
```

See https://github.com/puppetlabs/puppet/blob/7e210c334f0231e961b699c663c3efd4b32373bf/spec/integration/faces/config_spec.rb#L13 for an example

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,78 @@
...
+    install_fake_reboot_resource(agent)
+
+    step 'Run Puppet Apply'
+    on(agent, puppet('apply'), :stdin => dsc_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    # PuppetFakeResource always overwrites file at this path
+    on(agent, "cat /cygdrive/c/#{fake_name}", :acceptable_exit_codes => [0]) do |result|
```
Iristyle: Should we use a Powershell call with `EncodedCommand` here to avoid sending UTF-8 over SSH?

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,78 @@
...
+confine_block(:to, :platform => 'windows') do
+  agents.each do |agent|
+    step 'Apply Manifest to Remove File'
+    on(agent, puppet('apply'), :stdin => dsc_remove_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    # if this file exists, 'absent' didn't work
+    on(agent, "test -f /cygdrive/c/#{fake_name}", :acceptable_exit_codes => [1])
```
Iristyle: Should we use a Powershell call with EncodedCommand here to avoid sending UTF-8 over SSH?

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
ThoughtCrhyme: Yup, we can use variable bit width here.

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,85 @@
...
+confine_block(:to, :platform => 'windows') do
+  agents.each do |agent|
+    step 'Apply Manifest to Remove File'
+    on(agent, puppet('apply'), :stdin => dsc_remove_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    # if this file exists, 'absent' didn't work
+    on(agent, powershell("cat C://#{fake_name}", 'EncodedCommand' => true), :acceptable_exit_codes => [1])
```
Iristyle: I'm pretty sure `cat` in PowerShell doesn't set an exit code ... you'll want to make sure the contents are empty and `exit` the command yourself I think.

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    step 'Run Puppet Apply'
+    on(agent, puppet('apply'), :stdin => dsc_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    powershell_cmd = 'dir C:\\ | select Basename'
+    on(agent, powershell(powershell_cmd, 'EncodedCommand' => true)) do |result|
+      assert_match(/#{fake_name}/, result.stdout, 'File with correct UTF-8 characters was not present.')
```
Iristyle: This might be giving you a (false?) positive result, but I'm not sure it's quite right. You want to pass the UTF8 over the wire *to* the host using `EncodedCommand => true`. Something like this (going from memory, verify that I didn't screw anything up):

```ruby
powershell_cmd = "dir c:\\ | ? { $_.BaseName -eq '#{fake_name}' } | Measure-Object | Select Count"
```

This guarantees that PowerShell is taking Unicode input and using that for a comparison. Because we can't trust what's coming *back* from the remote host, then the assertion becomes that you got 1 matching result. Something like:

```ruby
assert_equal('1', result.stdout, 'File with correct UTF-8 characters was not present.')
```

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
Iristyle: Could also do a file count in the other example above, and verify there are 0 files by the given name.

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,85 @@
...
+MANIFEST
+
+# Teardown
+teardown do
+  confine_block(:to, :platform => 'windows') do
+    step 'Remove Test Artifacts'
+    agents.each do |agent|
+      uninstall_fake_reboot_resource(agent)
+    end
+    on(agents, "rm -rf /cygdrive/c/#{fake_name}")
```
Iristyle: We might want to do this a little differently to avoid the UTF-8 problem.

One solutions is to write these files to a randomized directory (ASCII characters only) and recursively delete it.

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,93 @@
...
+# different UTF-8 widths
+# 1-byte A
+# 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
+# 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+# 4-byte 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+MIXED_UTF8 = "A\u06FF\u16A0\u{2070E}" # Aۿᚠ𠜎
+
+# Manifest
+file_name = MIXED_UTF8
+file_path = (0...8).map { (65 + rand(26)).chr }.join
```
Iristyle: ^ this is fine, but `SecureRandom.uuid` does roughly the same thing ;0

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,93 @@
...
+    step 'Copy Test Type Wrappers'
+    install_fake_reboot_resource(agent)
+
+    step 'Run Puppet Apply'
+    on(agent, puppet('apply'), :stdin => dsc_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    powershell_cmd = "Write-Host (Get-ChildItem C:\\#{file_path} | Measure-Object ).Count;"
```
Iristyle: A little pedantic, but if we want to make sure we're matching the file by name we intended to create, we'd want:

```powershell
(Get-ChildItem C:\\#{file_path} -Filter '#{file_name}' | Measure-Object).Count
```

You may have run into errors doing something like `Get-ChildItem "C:\\#{file_path}\\#{file_name}"` but this will avoid that. Examples of executing code:

```
C:\source\> (Get-ChildItem .\locales\ -Filter 'config.foo' | Measure-Object).Count
0

C:\source\> (Get-ChildItem .\locales\ -Filter 'config.yaml' | Measure-Object).Count
1
```

With a randomized parent directory, there **shouldn't** be other files in there (unless there was a subtle test bug or a future test maintainer made a subtly incorrect change). I'm pointing this out for the specificity reason, but also because we don't need to use `EncodedCommand` with how you have the assert setup anymore. I think the `-Filter` is preferable, but if you feel strongly about the other, we should yank the `EncodedCommand` bit since `file_path` is only alpha-numeric.

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,93 @@
...
+MANIFEST
+
+# Teardown
+teardown do
+  confine_block(:to, :platform => 'windows') do
+    step 'Remove Test Artifacts'
+    agents.each do |agent|
+      uninstall_fake_reboot_resource(agent)
+    end
+    require "pry"; binding.pry
```
Iristyle: Call the zookeeper... we've got a wild `pry` on the loose

ThoughtCrhyme: lulz

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    step 'Run Puppet Apply'
+    on(agent, puppet('apply'), :stdin => dsc_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    powershell_cmd = "'Write-Host (Get-ChildItem C:\\#{file_path}  -Filter '#{file_name}' | Measure-Object ).Count;'"
+    on(agent, powershell(powershell_cmd)) do |result|
+      assert_match(/1/, result.stdout, 'File with correct UTF-8 characters was not present.')
```
Iristyle: What happens with multi-line output? Do we need to use anchors on the string (or maybe just `assert_equal`)?

tests/acceptance/tests/unicode/puppet_apply_utf8_file_name.rb
```diff
@@ -0,0 +1,92 @@
...
+    step 'Apply Manifest to Remove File'
+    on(agent, puppet('apply'), :stdin => dsc_remove_manifest, :acceptable_exit_codes => [0,2]) do |result|
+      assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+    end
+
+    step 'Verify Results'
+    # if the file count is greater than 0, 'absent' didn't work
+    powershell_cmd = "'Write-Host (Get-ChildItem C:\\#{file_path} -Filter '#{file_name}' | Measure-Object ).Count;'"
+    on(agent, powershell(powershell_cmd)) do |result|
+      assert_match(/0/, result.stdout, 'File with UTF-8 character name was not removed')
```
Iristyle: Same here regex vs `assert_equal`

PR #55, opened 2018-06-12
(MODULES-7197) Remove unused ensurable values

michaeltlombardi:
Prior to this commit the ensure property included an
option for users to specify `absent` on the resource
which may not actually match up with the DSC properties
and which follows the same code path as `present` in
any case. The ensure property also included `exists?`
as a valid value, even though setting it provided no
functionality for users.

This commit removes the code to support the `absent`
and `exists?` values in the type and the `absent`
value in the provider. The `exists?` codepath is still
called before puppet runs the `create` code path, and
only runs `create` if the returned value from `exists?`
is `false`.

Further, this commit removes the ensure tests from the
acceptance test suite. These tests verified functionality
when the dsc and puppet `ensure` properties were in
conflict. As of this commit there is only one valid value
for `ensure` on the puppet resource so these tests are no
longer needed.

Finally, this commit adds documentation to the `ensure`
property and some comments to clarify why we did not
change the name of the ensurable value (reporting).

A future commit which addresses the reporting problems
may also allow us to update the ensurable value from
'present' to something like 'invoke' which is more
semantically accurate.


README.md
```diff
@@ -287,6 +287,11 @@ The `dsc` type allows specifying any DSC Resource declaration as a minimal Puppe
 
 The name of the declaration. This has no affect on the DSC Resource declaration and is not used by the DSC Resource.
 
+#### ensure
+
+A necessary property to cause puppet to invoke the dsc resource if required.
```
glennsarti: Not keen on this.  Perhaps

```
An optional property that specifies that the DSC resource should be invoked.  This property has only one value of `present`.  This property does not need be be set in manifests.
```

lib/puppet/type/dsc.rb
```diff
@@ -5,9 +5,17 @@
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_lite/dsc_type_helpers'
 
   ensurable do
-    newvalue(:exists?) { provider.exists? }
+    desc <<-HERE
+    A necessary property to cause puppet to invoke the dsc resource if required.
```
glennsarti: As per comment in README

glennsarti:
Trap is clean

PR #57, opened 2018-06-20
(MODULES-6572) Test Suite Refactors

michaeltlombardi:
These commits refactor the acceptance tests in various minor ways; this is all tech-debt management.


michaeltlombardi:
I have some commits to cleanup.

michaeltlombardi:
Kicked off an adhoc run since this touches all of the acceptance suite.

michaeltlombardi:
[Adhoc has passed](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/dsc_lite/job/forge-windows_puppetlabs-dsc_lite_intn-sys_full-pe-previous-adhoc/10/)

tests/lib/dsc_utils.rb
```diff
@@ -176,3 +176,7 @@ def assert_reboot_pending(host)
     end
   end
 end
+
+def windows_agents
+  agents.select { |agent| agent['platform'].include?('windows') }
+end
```
glennsarti: Missing trailing LF

tests/lib/dsc_utils.rb
```diff
@@ -123,13 +123,13 @@ def get_fake_reboot_resource_install_path(usage = :manifest)
...
   # Init
-  dsc_resource_target_path = 'lib/puppet_x/dsc_resources'
+  dsc_resource_target_path = 'lib/puppet_x/dsc_resources/PuppetFakeResource'
   puppet_type_target_path = 'lib/puppet/type'
 
   step 'Determine Correct DSC Module Path'
   dsc_module_path = locate_dsc_module(host)
-  dsc_resource_path = "#{dsc_module_path}/#{dsc_resource_target_path}/PuppetFakeResource"
-  dsc_resource_path2 = "#{dsc_module_path}/#{dsc_resource_target_path}/PuppetFakeResource2"
+  dsc_resource_path = "#{dsc_module_path}/#{dsc_resource_target_path}1.0"
```
Iristyle: Are there slashes missing here?

tests/acceptance/tests/reboot/reboot_autonotify.rb
```diff
@@ -3,15 +3,21 @@
...
+  dsc_resource_module => '#{installed_path}/1.0',
+  dsc_resource_properties => {
+    importantstuff  => 'reboot',
+    requirereboot   => true,
+  }
+}
 reboot { 'dsc_reboot':
   when => pending
 }
-dsc_puppetfakeresource { 'reboot_test':
```
Iristyle: We were still using the old DSC style wrappers here because the new resources didn't support integration with reboots yet.

If the tests are passing now, this is a bit concerning (unless reboot support has been added to `dsc_lite` now?).

glennsarti: https://github.com/puppetlabs/puppetlabs-dsc_lite/blame/master/lib/puppet/provider/base_dsc_lite/powershell.rb#L113 ?

Iristyle: That's not the bit I'm referring to unfortunately. 
For that to do anything, the graph needs to be built. All the generated types in the old module contain code like this to do that - https://github.com/puppetlabs/puppetlabs-dsc/blob/master/lib/puppet/type/dsc_archive.rb#L176

The new module doesn't have an equivalent yet to my knowledge.

I'm on mobile at the moment and haven't double checked the test intent, but I thought at least some of it needed the graph connected properly.

Note that the old code still doesn't do "the right thing" / what we originally intended, as it builds the graph wrong.

michaeltlombardi: If that's the case, it sounds like these tests shouldn't be running (if they're not testing `dsc_lite` resources) or that we need to add a ticket to explore why they suddenly work. I'm happy to add a skip for now if we think that's a good idea, but as the tests are passing it seems like we want to file a ticket instead and just merge.

michaeltlombardi: It looks like the reboot functionality is scheduled for 1.1.0 epic:

glennsarti:
Passed adhoc.

```
21:19:03               - Test Case Summary for suite 'tests' -
21:19:03        Total Suite Time: 242.85 seconds
21:19:03       Average Test Time: 16.19 seconds
21:19:03               Attempted: 15
21:19:03                  Passed: 10
21:19:03                  Failed: 0
21:19:03                 Errored: 0
21:19:03                 Skipped: 0
21:19:03                 Pending: 5
21:19:03                   Total: 15
21:19:03 
21:19:03       - Specific Test Case Status -
```

tests/acceptance/tests/reboot/reboot_autonotify.rb
Iristyle: Yeah, performing a skip seems like the right thing to do for now.  We can defer debugging of those tests until the `reboot` functionality is implemented. The helper `assert_reboot_pending` at https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/tests/lib/dsc_utils.rb#L171-L176 has this code:

```ruby
on(host, 'shutdown /a', :accept_all_exit_codes => true) do |result|
  assert(0 == result.exit_code, 'Expected reboot is not pending!')
end
```

With no pending reboot, we should get an exit code like `1116`:

```
C:\Users\Administrator>shutdown /a
Unable to abort the system shutdown because no shutdown was in progress.(1116)

C:\Users\Administrator>echo %errorlevel%
1116
```

Or similar under Cygwin:

```
Administrator@ebjd161bkm27mi4 ~
$ shutdown /a
Unable to abort the system shutdown because no shutdown was in progress.(1116)

Administrator@ebjd161bkm27mi4 ~
$ echo $?
92
```

That part all makes sense to me, but I'm wondering if something on these systems is always making the `reboot` resource think that there is a `pending` reboot? All the manifests have some flavor of this:

```puppet
reboot { 'dsc_reboot':
  when => pending
}
```

That isn't enough to discern a `dsc` based reboot, so could generate false positives.  Interestingly one of the tests removes a registry key for exactly that reason at https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/tests/acceptance/tests/reboot/reboot_no_invoke.rb#L40-L42: 

```ruby
# Workaround for https://tickets.puppetlabs.com/browse/IMAGES-539
step 'Remove PendingFileRenameOperations registry key'
on(agent, 'reg delete "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager" /v PendingFileRenameOperations /f', :accept_all_exit_codes => true)
```

But only one of the 5 tests does that, which seems bad.

* [reboot_autonotify.rb](https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/tests/acceptance/tests/reboot/reboot_autonotify.rb) - I'd expect this test fails now, so it's puzzling if it does not. But perhaps it's running first and the reg key is still present.
* [reboot_explicit.rb](https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/tests/acceptance/tests/reboot/reboot_explicit.rb) - Arguably this isn't needed because it's testing a Puppet `notify`. This test would still pass.
* [reboot_no_invoke.rb](https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/tests/acceptance/tests/reboot/reboot_no_invoke.rb) - This verifies *no* reboot is pending, which should still pass.
* [reboot_warning.rb](https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/tests/acceptance/tests/reboot/reboot_warning.rb) - This verifies *no* reboot is pending, which should still pass.
* [negative/reboot_graph_cycle.rb](https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/tests/acceptance/tests/reboot/negative/reboot_graph_cycle.rb) - This verifies *no* reboot is pending, which should still pass.

So my guess here is that we have some dependencies on test ordering.




For additional reference, the DSC templates have the code https://github.com/puppetlabs/puppetlabs-dsc/blob/master/build/dsc/templates/dsc_type.rb.erb#L258-L261 to wire up the graph:

```ruby
def builddepends
  pending_relations = super()
  PuppetX::Dsc::TypeHelpers.ensure_reboot_relationship(self, pending_relations)
end
```

MODULES-6569 will have us adding something like that in the actual `dsc` type - https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/lib/puppet/type/dsc.rb#L76. Not sure yet how that will shake out, given the current behavior in the `dsc` module is not quite what we actually want.

Iristyle:
My questions are resolved for now - but since I'm not as close on this one, will let @glennsarti / @ThoughtCrhyme merge this one.

PR #58, opened 2018-06-21
(MODULES-7253) Replace verbose parameter names

michaeltlombardi:
Prior to this commit the dsc resource used the following long,
slightly repetitive names for the parameters:

- `dsc_resource_name`
- `dsc_resource_module`
- `dsc_resource_properties`

This commit replaces those names with the following:

- `resource_name`
- `module`
- `properties`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Unreleased
 
+### Changed
+
+- **BREAKING:**  Renamed all parameters for resource declarations, will require manifest modifications. [(MODULES-7253)](https://tickets.puppetlabs.com/browse/MODULES-7253).
+
+
 ## 2018-06-07 - Unsupported Release 0.4.0
 
 ### Summary
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ See [known issues](#known-issues) for troubleshooting setup.
 
 ## Usage
 
-The generic `dsc` type is a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. You can use a DSC Resource by supplying the same properties you would set in a DSC Configuration script inside the `dsc_resource_properties` parameter. For most use cases the `dsc_resource_properties` parameter accepts the same structure as the PowerShell syntax, with the substitution of Puppet syntax for arrays, hashes and other data structures.
+The generic `dsc` type is a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. You can use a DSC Resource by supplying the same properties you would set in a DSC Configuration script inside the `properties` parameter. For most use cases the `properties` parameter accepts the same structure as the PowerShell syntax, with the substitution of Puppet syntax for arrays, hashes and other data structures.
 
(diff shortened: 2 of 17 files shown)


michaeltlombardi:
So, I went to go and alias the values backwards... and discovered this isn't actually a capability inside of puppet as far as I can see.

Advice I got from the modules team was to consider following the pattern from the firewall resource:

https://github.com/puppetlabs/puppetlabs-firewall/blob/9a17911503d0b93016349a086842a03a6256f795/lib/puppet/type/firewall.rb#L267-L331

Notice that here they copied the original parameter over to the new name. The old parameter has a validate block that raises a deprecation warning. I can implement this if this is the direction we want to take.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 ## Unreleased
 
+### Changed
+
+- **BREAKING:** Replaced the existing parameters for the resource declarations as per below [(MODULES-7253)](https://tickets.puppetlabs.com/browse/MODULES-7253).
```
jpogran: Prefer `renamed` rather than `replaced`.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+  Note that these changes _will_ require updating your manifests.
```
jpogran: This format breaks what we have been doing so far with the CHANGELOG. As you have it now, its `<Severity>: Message <Ticket link> Another Message`. Up until now we had `<Severity>: Short Message <Ticket link>`.

I think this information is better served in the ticket and this line kept short:

**BREAKING:** Renamed all parameters for resource declarations, will require manifest modifications. [(MODULES-7253)](https://tickets.puppetlabs.com/browse/MODULES-7253)

michaeltlombardi: In that case, we should link to the section of the docs that includes reference for these parameters.

michaeltlombardi: On the other hand, I don't actually know if that link will be available on the forge. I suppose we'll just have to trust users to change tabs on their own. :)

Iristyle:
I don't think we have to address in this PR, but what about the nested `dsc_type` and `dsc_properties` for embedded `CimInstance` hashes? I'm less concerned about those as they should be "rarer"... and we probably do need some "namespacing" to avoid collisions. But wondering if this has been discussed?

michaeltlombardi:
@Iristyle ah, crap, no, that topic didn't come up at all in this ticket until now.

michaeltlombardi:
After talking it over with @glennsarti, filed [MODULES-7352](https://tickets.puppetlabs.com/browse/MODULES-7352) for further discussion and implementation.

jpogran:
I didn't suggest those to change because they need to be unique enough to [find](https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/master/lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb#L61-L76) in the hash. They can't be something 'normal' as it could conflict with parameters used by DSC Resources.

PR #59, opened 2018-06-29
(MODULES-7179) Change created to invoked

RandomNoun7:
This commit changes the message that Puppet returns when dsc_lite
resources are run from "created" to "invoked". This is because at the
moment, when we invoke a dsc_lite resource, Puppet does not know if the
resource we want to manage has actually been created, or if in reality
it has simply been modified, or even deleted. Changing to invoked is
more ambiguous, but accurately reflects the idea that we don't truly
know what operation dsc has accomplished for us.


RandomNoun7:
@ThoughtCrhyme I think changing the verbage in those two tests will be fine as far as testing the change in this PR, just because if someone does revert to returning 'changed', these tests will definitely fail, and there aren't really any additional useful code paths to worry about.

PR #63, opened 2018-07-06
(MAINT) make unicode tests conform to previous dsc tests

ThoughtCrhyme:


spec/spec_helper_acceptance.rb
```diff
@@ -32,3 +32,17 @@ def beaker_opts
...
+
+# Due to IMAGES-825, sometimes we can't create folders from within Cygwin, but instead
+# need to break out to native Windows binaries to create them.  This creates files and
+# folders with a correct and sane DACL
+def create_remote_windows_directory(hosts, windows_path)
+  windows_path = windows_path.gsub('\\','\\\\\\')
+  on(hosts, "cmd.exe /c mkdir #{windows_path}", :accept_all_exit_codes => true)
+end
+
+def create_remote_windows_file(hosts, windows_path, content)
```
RandomNoun7: It looks to me like these functions are defined and then never used. Is this just proactive addition because of a known issue?

PR #69, opened 2018-07-16
(MODULES-7178) Report types dynamically

Iristyle:
- [x] ~Determine if new tests can / should be added~ basic specs added
 - [x] ~~Should `resource_name` be canonicalized? i.e. Does `FiLe` show up differently than `file`... do we care? (Report logic already calls `capitalize` and the report looks fine, but should the internal tracking maintain state differently?)~~ Canonicalized by calling `downcase` on the `resource_name` property so that case differences don't matter internally.
 - [x] Make sure nothing else weird happens in the system when the type reports as `:dsc_foo` instead of `:dsc`. ~~Are there any other dependencies on the symbol `:dsc`?~~ There are no other usages of `:dsc`
 - [x] ~~Make sure that behavior on an older PE install is reasonable (without the PUP-8746 / https://github.com/puppetlabs/puppet/commit/28240e5e300e8548be33142e108cf706a788bd74 fix) - it may end up behaving partially correctly if metrics summary is out of sync with actual resource types~~ Screenshots / explanation below. Behavior is OK and as expected.
 - [x] ~~Test in conjunction with the DSC resource that may refer to the same types (should we use `DscLite_` instead of `Dsc_` for the prefix?)~~ Vetted this with @michaeltlombardi - didn't include screenshots, but it was necessary to prefix with `dsc_lite_` to avoid collisions with DSC module

- Prior to this commit, all types would report the `dsc` type as their
   identity. This is problematic for the sake of reporting as every type
   of resource is grouped together under `dsc`, rather than the specific
   `resource_name` modeled in a manifest.

   This new approach allows for types to report under something more
   useful that indicates the original DSC resource type specified in the
   resource_name attribute.

   For instance, given a manifest with the following resources:

```puppet
   dsc { 'sample_file':
     resource_name     => 'File',
     module            => 'PSDesiredStateConfiguration',
     properties        => {
       ensure          => 'present',
       contents        => 'hello',
       destinationpath => 'c:\\test.txt',
     }
  }

  dsc { 'lmhosts':
     resource_name     => 'Service',
     module            => 'PSDesiredStateConfiguration',
     properties        => {
       name            => 'lmhosts',
       ensure          => 'present',
       state           => 'running',
     }
  }
```

  The report now shows:

  * Dsc_lite_file[sample_file]    Unchanged
  * Dsc_lite_service[lmhosts]     Unchanged

  Instead of what it previously reported:

  * Dsc[sample_file]              Unchanged
  * Dsc[lmhosts]                  Unchanged

 - Selection of the PowerShell code was previously determined by type,
   but given it has changed to being specified as :Dsc_lite_XXXX, its
   not as easy to do that anymore. Instead, add a simple boolean
   property to the generic dsc wrapper that may be checked instead.

 - Note that parameter validation for `module`, `resource_name` and
   properties does not currently work, but will be fixed in another
   ticket by adding a global validator. The `type` method performs some
   additional guarding around usage of the `resource_name` parameter
   for 2 reasons:

   * It's unclear if the parser or other logic in Puppet uses the type
     name without the resource being instantiated fully with values from
     a manifest
   * Due to the incorrect validation, the PE console errors with the
     following message when `resource_name` is left unspecified:

   undefined method `value` for nil:NilClass

### DIFF
--- a/lib/puppet/type/dsc.rb
+++ b/lib/puppet/type/dsc.rb
@@ -4,6 +4,20 @@ Puppet::Type.newtype(:dsc) do
   require Pathname.new(__FILE__).dirname + '../../' + 'puppet/type/base_dsc_lite'
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_lite/dsc_type_helpers'
 
+  def type
+    # in the event this value is consumed early in the lifecycle / before
+    # parameters have been populated, use 'unspecified'
+    name = ( parameters[:resource_name].nil? || parameters[:resource_name].value.nil? || (parameters[:resource_name].value =~ /\W/) ) ?
+      'unspecified' :
+      # but for the sake of catalog eval / report status, masquerade as a different type
+      parameters[:resource_name].value.downcase
+    "Dsc_lite_#{name}".to_sym
+  end
+
+  def generic_dsc
+    true
+  end
+
   ensurable do
     desc <<-HERE
     An optional property that specifies that the DSC resource should be invoked.
--- a/lib/puppet/provider/base_dsc_lite/powershell.rb
+++ b/lib/puppet/provider/base_dsc_lite/powershell.rb
@@ -169,7 +169,7 @@ EOT
   def self.ps_script_content(mode, resource, provider)
     dsc_invoke_method = mode
     @param_hash = resource
-    template_name = resource.type == :dsc ?
+    template_name = resource.generic_dsc ?
       '/invoke_generic_dsc_resource.ps1.erb' :
       '/invoke_dsc_resource.ps1.erb'
     file = File.new(template_path + template_name, :encoding => Encoding::UTF_8)
--- a/spec/unit/puppet/type/dsc_spec.rb
+++ b/spec/unit/puppet/type/dsc_spec.rb
@@ -8,6 +8,22 @@ describe Puppet::Type.type(:dsc) do
 
   it { is_expected.to be_a_kind_of Puppet::Type::Dsc }
 
+  describe "type" do
+    it "should be built dynamically from parameter :resource_name" do
+      resource[:resource_name] = 'foo'
+      expect(resource.type).to eq(:Dsc_lite_foo)
+    end
+
+    it "should return Dsc_lite_unspecified given a missing :resource_name" do
+      expect(resource.type).to eq(:Dsc_lite_unspecified)
+    end
+
+    it "should return Dsc_lite_unspecified given a :resource_name of ' '" do
+      resource[:resource_name] = ' '
+      expect(resource.type).to eq(:Dsc_lite_unspecified)
+    end
+  end
+
   describe "parameter :name" do
     subject { resource.parameters[:name] }
 


Iristyle:
Given a site.pp on a 2018.1.2 master (which contains the patch for PUP-8746 at https://github.com/puppetlabs/puppet/pull/6838, which is part of Puppet 5.5.2 and 5.5.3)

```
node default {
      dsc {'8c8ffeaf-ddee-46a7-8bc4-0749ad256be4':
      resource_name => 'file',
      module => 'PSDesiredStateConfiguration',
      properties => {
        ensure          => 'absent',
        contents  => '36b5ccba-84fe-470d-a064-8d12c671918b',
        destinationpath => 'C:\8c8ffeaf-ddee-46a7-8bc4-0749ad256be4'
      }
    }

}

dsc {'sample_foo':
  resource_name     => 'service',
  module            => 'PSDesiredStateConfiguration',
  properties        => {
    ensure          => 'present',
    state           => 'running',
    name            => 'lmhosts',
  }
}

dsc {'sample_file2':
  resource_name     => 'File',
  module            => 'PSDesiredStateConfiguration',
  properties        => {
    ensure          => 'present',
    contents        => 'hello',
    destinationpath => 'c:\\test2.txt',
  }
}
```

The behavior is as desired:

screen shot 2018-07-16 at 4 28 47 pm
screen shot 2018-07-16 at 4 29 03 pm


To repro, I provisioned using the following procedure:

```
GEM_SOURCE=http://rubygems.delivery.puppetlabs.net BEAKER_keyfile=~/.ssh/id_rsa-acceptance BEAKER_debug=true BEAKER_destroy=no BEAKER_PE_DIR=http://pe-releases.puppetlabs.lan/2018.1.2/ BEAKER_PE_VER=2018.1.2 BEAKER_setfile=2012r2.yaml TEST_FRAMEWORK=beaker-rspec BEAKER_TESTMODE=agent PUPPET_INSTALL_TYPE=pe bundle exec rake beaker
```

Once provisioned, I modified the `site.pp` on the master using `vi /etc/puppetlabs/code/environments/production/manifests/site.pp`, adding the above code. The suite installs the module to `/etc/puppetlabs/code/modules/dsc_lite/`, where it may be further edited.  To get pry-byebug / pry-stack_explorer on the Windows node for debugging purposes, I did the following:

* Installed Ruby 2.4 installer from https://github.com/oneclick/rubyinstaller2/releases/download/rubyinstaller-2.4.4-2/rubyinstaller-devkit-2.4.4-2-x64.exe
* Made sure that devkit / msys were also installed
* Performed an install of the relevant gems using `gem install pry-byebug pry-stack_explorer --no-ri --no-rdoc`
* Copied the built copies of the gems from `C:\Ruby24-x64\lib\ruby\gems\2.4.0` to the Puppet vendored Ruby at `C:\Program Files\Puppet Labs\Puppet\sys\ruby\lib\ruby\gems\2.4.0`. I copied the following installed / built bits from the directories `cache`, `gems` and `specifications`:
    * `binding_of_caller`
    * `byebug`
    * `coderay`
    * `debug_inspector`
    * `method_source`
    * `pry`
    * `pry-byebug`
    * `pry-stack_explorer`
* I then opened the `Start Command Prompt with Puppet`
* From inside that Ruby environment, I installed one additional gem to get pry to work properly - `gem install rb-readline --no-ri --no-rdoc`

lib/puppet/type/dsc.rb
```diff
@@ -4,6 +4,15 @@
   require Pathname.new(__FILE__).dirname + '../../' + 'puppet/type/base_dsc_lite'
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_lite/dsc_type_helpers'
 
+  def type
+    # for the sake of report status, masquerade as a different type
+    ("Dsc_" + parameters[:resource_name].value).to_sym
```
glennsarti: Huh... TIL ... I thought you needed to use `.intern`

glennsarti:
Those tests you mentioned seem to cover most of where I think the "unintended consequences" would come about.  I would also consider adding in a resource name collision on purpose to see what happens.  e.g.

```
dsc_file { 'sample_file1':
....
}

dsc {'sample_file2':
  resource_name     => 'File',
...
}
```

And possibly two resources with the same title
```
dsc_file { 'collide':
....
}

dsc {'collide':
  resource_name     => 'File',
...
}
```
Both in an agent only and a master/agent setup.

To see what the fall out would be.

lib/puppet/type/dsc.rb
Iristyle: Changing to `DscLite_` as the prefix to prevent collisions with existing DSC module.

Iristyle: To canonicalize the type name, I think we should do a `parameters[:resource_name].value.capitalize`

lib/puppet/type/dsc.rb
```diff
@@ -4,6 +4,20 @@
   require Pathname.new(__FILE__).dirname + '../../' + 'puppet/type/base_dsc_lite'
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_lite/dsc_type_helpers'
 
+  def type
+    # in the event this value is consumed early in the lifecycle / before
+    # parameters have been populated, use 'unspecified'
+    name = ( parameters[:resource_name].nil? || parameters[:resource_name].value.nil? || parameters[:resource_name].value.match?(/\W/) ) ?
```
Iristyle: This extra check can be removed once https://tickets.puppetlabs.com/browse/MODULES-7485 is fixed.

PR #71, opened 2018-07-19
(MODULES-7485) Fix validation for required parameters

jpogran:
This commit uses global type validation to require the `module`,
`resources_name`, and `properties`, instead of `isrequired`. `isrequired`
does not work ( see https://projects.puppetlabs.com/issues/4049 / https://tickets.puppetlabs.com/browse/PUP-1591 )

### DIFF
--- a/lib/puppet/type/dsc.rb
+++ b/lib/puppet/type/dsc.rb
@@ -5,12 +5,7 @@ Puppet::Type.newtype(:dsc) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_lite/dsc_type_helpers'
 
   def type
-    # in the event this value is consumed early in the lifecycle / before
-    # parameters have been populated, use 'unspecified'
-    name = ( parameters[:resource_name].nil? || parameters[:resource_name].value.nil? || (parameters[:resource_name].value =~ /\W/) ) ?
-      'unspecified' :
-      # but for the sake of catalog eval / report status, masquerade as a different type
-      parameters[:resource_name].value.downcase
+    name = parameters[:resource_name].value.downcase
     "Dsc_lite_#{name}".to_sym
   end
 
@@ -56,7 +51,6 @@ HERE
 
   newparam(:resource_name) do
     desc "DSC Resource Name"
-    isrequired
     validate do |value|
       if value.nil? or value.empty?
         raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
@@ -67,7 +61,6 @@ HERE
 
   newparam(:module) do
     desc "DSC Resource Module"
-    isrequired
     validate do |value|
       if value.nil? or value.empty?
         raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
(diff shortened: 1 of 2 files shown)


spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -13,15 +20,6 @@
       resource[:resource_name] = 'foo'
       expect(resource.type).to eq(:Dsc_lite_foo)
     end
-
-    it "should return Dsc_lite_unspecified given a missing :resource_name" do
```
jpogran: @Iristyle these tests were added from the reporting changes, but as I explain in the comment this situation will never happen because we now require resource_name and fail if it's not specified. If left in, calls to `resource.type` always return the correct name

Iristyle: Yup, also noted that we can remove some of the guards in `type` now - see https://github.com/puppetlabs/puppetlabs-dsc_lite/pull/69#discussion_r203219633

lib/puppet/type/dsc.rb
```diff
@@ -88,14 +86,22 @@ def change_to_s(currentvalue, newvalue)
...
     validate do |value|
       if value.nil? or value.empty?
         raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
       end
       fail "#{self.name.to_s} should be a Hash" unless value.is_a? ::Hash
     end
   end
+
+  validate do
+    raise ArgumentError, 'dsc: resource_name is required' unless self[:resource_name]
```
michaeltlombardi: Do we also want to guard these against people supplying the values as `nil` or `''`?

michaeltlombardi: Whelp, ignore me, I can't read.

_Looks at the validate blocks for the parameters_

Iristyle: Yeah, its a little confusing given we have multiple validations that run at different points in the lifecycle. I think it's still worth seeing (even if it ends up just being for academic reasons) what message we see in the console with `:resource_name` set to `''` or `undef` - just to verify the UX.

Iristyle:
I think we probably want to merge the commits together, given the first commit doesn't pass tests.

lib/puppet/type/dsc.rb
```diff
@@ -98,6 +98,8 @@ def change_to_s(currentvalue, newvalue)
     raise ArgumentError, 'dsc: resource_name is required' unless self[:resource_name]
     raise ArgumentError, 'dsc: module is required' unless self[:module]
     raise ArgumentError, 'dsc: properties is required' unless self[:properties]
+
+    provider.validate if provider.respond_to?(:validate)
```
Iristyle: Our provider doesn't have a `validate` - was this an idiom you saw used elsewhere?

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -20,15 +20,6 @@
       resource[:resource_name] = 'foo'
```
Iristyle: Think we can also remove this line, as it's already being set by default in the call at line 7.

lib/puppet/type/dsc.rb
```diff
@@ -88,14 +81,22 @@ def change_to_s(currentvalue, newvalue)
...
       fail "#{self.name.to_s} should be a Hash" unless value.is_a? ::Hash
     end
   end
+
+  validate do
+    raise ArgumentError, 'dsc: resource_name is required' unless self[:resource_name]
+    raise ArgumentError, 'dsc: module is required' unless self[:module]
+    raise ArgumentError, 'dsc: properties is required' unless self[:properties]
+
+    provider.validate if provider.respond_to?(:validate)
```
Iristyle: Only type / provider I found in Puppet that does this is the `file` type, for the sake of the Windows file provider.

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/file/windows.rb#L83-L87

In the general sense here I don't think it's necessary as we don't have a provider that has a specific `validate` (we don't have multiple providers, which is the case for `file`), and I don't expect we will in the future.

PR #76, opened 2018-08-08
Add an example of identifying available parameters

RandomNoun7:
With native Puppet types (including the `dsc_*` types), we're used to being able use `puppet describe` to explore the types and discover available parameters. When porting from `dsc` to `dsc_lite`, the lack of that ability might be a sticking point, especially for admins who primarily administer Linux boxes and aren't comfortable enough with PowerShell to know how to do this on their own.

To mitigate that, we'll just show an example in the REAME.

(Can someone with an actual Windows box correct this example to be consistent with the `WindowsFeature` code snippets that follow?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,29 @@ See [known issues](#known-issues) for troubleshooting setup.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


RandomNoun7:
This content is feedback from @binford2k.

PR #77, opened 2018-08-13
(MODULES-7556) Note added to README

clairecadman:
This commit adds a note about dsc_lite being for advanced users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@
 
 The Puppet dsc_lite module allows you to manage target nodes using Windows PowerShell DSC (Desired State Configuration) Resources.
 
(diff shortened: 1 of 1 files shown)


RandomNoun7:
Are we worried that asking for experience with DSC might put some people off of trying this thing out? Would the revision in the first commit asking for the slightly lower bar of some knowledge rather than experience perhaps be slightly better? I'm fine with this version if we are really concerned that people might get frustrated and drop it entirely.

murdok5:
@RandomNoun7 even though it was merged, I think I am comfortable with the change, since if they are "new" they should start with the DSC module, not DSC_Lite.  Right?

RandomNoun7:
@murdok5 I kind of think that once we have a public webinar up that we can refer people to that and that it won't be unreasonable for people to jump into DSC_Lite with that base layer of knowledge. I think that once we have up and available that we might edit this again and refer people to the webinar if they are in any way unclear on any of the terminology.

PR #85, opened 2018-10-10
(MODULES-8602) Resolved declaration conflict between DSC/DSC_Lite

AnthonieSmitASR:
I have changed the name of the variable 'COMMAND_TIMEOUT' to 'DSC_LITE_COMMAND_TIMEOUT'. The reason for this change is that this resolves a double declaration issue when using both DSC and DSC_Lite modules on the same Puppet master.


michaeltlombardi:
Passed adhoc, tests pass, LGTM!

PR #91, opened 2018-12-06
(MODULES-8175) Add safety to new-pscredential helper

michaeltlombardi:
Prior to this commit the helper function `new-pscredential`
in the template used to invoke DSC resources attempted to
cast an incoming value for password to a string. When passed
a hashtable, it was unable to extract the password for use
and instead set the password as the typename for the object.

This happened when the catalog was compiled on a Puppet 5
master to apply to a Puppet 6 agent or when compiled on a
Puppet 6 master for a Puppet 5 agent and specifying the
password using the sensitive datatype.

This problem was introduced in MODULES-8597 when the tags
for rich data serialization were changed from `__pcore_type_`
and `__pcore_value__` to the shorter `__ptype` and `__pvalue`.
After this change, the master and agent both expect different
keys for the sensitive data type, which is why this problem
only shows when the master and agent are mismatched versions.

This commit modifies the `new-pscredential` helper function
to include handling for both strings and hashtables, and
will extract the password if possible and error loudly if
not.

### DIFF
--- a/lib/puppet/provider/base_dsc_lite/invoke_dsc_resource.ps1.erb
+++ b/lib/puppet/provider/base_dsc_lite/invoke_dsc_resource.ps1.erb
@@ -12,9 +12,27 @@ function new-pscredential
 
     [parameter(Mandatory=$true,
       ValueFromPipelineByPropertyName=$true)]
-    [string]
+    [object]
     $password
   )
+
+  Switch ($password.GetType().FullName) {
+    'System.String' {
+      [string]$password = $password
+    }
+
+    'System.Collections.Hashtable' {
+      If ($password.__pcore_type__ -eq 'Sensitive') {
+        [string]$password = $password.__pcore_value__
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/base_dsc_lite/invoke_dsc_resource.ps1.erb
```diff
@@ -12,10 +12,28 @@ function new-pscredential
 
     [parameter(Mandatory=$true,
       ValueFromPipelineByPropertyName=$true)]
-    [string]
+    [object]
     $password
   )
 
+  Switch ($password.GetType().FullName) {
```
Iristyle: Hmmmm - I'm a little confused on why we're doing all of this special handling on the PowerShell side instead of just munging the value consistently on the Ruby side?

Iristyle:
This will be reverted in favor of #92

PR #92, opened 2018-12-10
(MODULES-8175) Munge Properties Hash for Sensitive

michaeltlombardi:
Prior to this commit the PowerShell template code included a workaround
which munged a malformed hash (which should have been a sensitive value)
in the `New-PSCredential` method.

This commit adds a deserialization method and uses it in the munging
of the properties hash itself, obviating the need to handle this in
the PowerShell template code.

This also handles sensitive values for things other than the
MSFT_Credential DSC type, which the prior implementation did not.

### DIFF
--- a/lib/puppet/type/dsc.rb
+++ b/lib/puppet/type/dsc.rb
@@ -87,6 +87,10 @@ HERE
       end
       fail "#{self.name.to_s} should be a Hash" unless value.is_a? ::Hash
     end
+
+    munge do |value|
+      PuppetX::DscLite::TypeHelpers.munge_sensitive_hash(value)
+    end
   end
 
   validate do
--- a/lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
+++ b/lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
@@ -8,6 +8,21 @@ module PuppetX
         raise ArgumentError.new "invalid value: #{value}"
       end
 
+      # This is used to recursively search the hash to find sensitive data
+      # and ensure it is properly represented.
+      # It will modify the values in the hash passed to it.
+      def self.munge_sensitive_hash(h)
+        if h.is_a?(Hash) && h['__pcore_type__'] == 'Sensitive' && h.key?('__pcore_value__')
+          return Puppet::Pops::Types::PSensitiveType::Sensitive.new(h['__pcore_value__'])
+        elsif h.is_a?(Hash) && h['__ptype'] == 'Sensitive' && h.key?('__pvalue')
+          return Puppet::Pops::Types::PSensitiveType::Sensitive.new(h['__pvalue'])
+        else
+          h.each_pair do |key, value|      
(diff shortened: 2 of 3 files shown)


michaeltlombardi:
@hlindberg, this is related to the mismatch in keys for Sensitive where the master and agent are of mismatched versions (Puppet 5 master / 6 agent and vice versa).

We were able to workaround the problem in the module, but not sure if this could/should be fixed in the agent code?

michaeltlombardi:
Note: The changes have passed ad-hoc.

hlindberg:
I doubt this can be fixed as it it is a matter of having rich-data support or not, and if the catalog stringifies rich data or not. The work around handles one case, when the catalog contains rich data but the agent does not deserialize it (and ends up with the raw __pcore_xxx hashes). Note that later version of puppet uses `__ptype` and `__pvalue` instead of the longer tags - but then it should also be using the correct deserialization, so should not be a problem.

michaeltlombardi:
@hlindberg We found this to happen with newer master and older agent, _and_ with older master / newer agent - how likely is this to pop up in other modules that use sensitive when there's a mismatch?

This code _should_ handle both `__pcore_*` and `__p*` hashes, I think.

hlindberg:
@michaeltlombardi I doubt it is common for modules to have nested structures with sensitive values inside them. We would have heard more complaints if this was common.

(I did indeed miss that `__p*` cases were in there...)

The "newer master, older agents" case worries me since there you you get stringification where the older master probably issued warnings and sent rich data. The other way around isn't really a supported case (new agents/old master) even if some combinations just happens to be compatible.

lib/puppet/provider/base_dsc_lite/invoke_dsc_resource.ps1.erb
```diff
@@ -12,28 +12,10 @@ function new-pscredential
 
     [parameter(Mandatory=$true,
       ValueFromPipelineByPropertyName=$true)]
-    [object]
+    [string]
```
Iristyle: Rather than doing this work here, let's just revert #91

Iristyle:
Once we've run through the various combos / documented, and made sure this works in all cases, I think we can merge.

Since I think I'll be OoO tomorrow, was hoping you might have a look as well @joshcooper. Thanks!

michaeltlombardi:
So far I've tested against:

- Master 6x
  - [x] Agent 5x (rich_data false)
  - [x] Agent 5x (rich_data true)
  - [x] Agent 5x (apply)
  - [x] Agent 6x (rich_data false)
  - [x] Agent 6x (rich_data true)
  - [x] Agent 6x (apply)
- Master 5x
  - [x] Agent 5x (rich_data false)
  - [x] Agent 5x (rich_data true)
  - [x] Agent 5x (apply)
  - [x] Agent 6x (rich_data false)
  - [x] Agent 6x (rich_data true)
  - [x] Agent 6x (apply)

Anything else we should add to the matrix? I tried toggling rich_data off on the servers but it didn't seem to make a difference.

PR #94, opened 2018-12-20
(MODULES-8397) PDK Sync module to fix Travis

RandomNoun7:
This change syncs the module with the latest version of the pdk template
files. The primary objective is to ensure that the .travis.yml file
is updated. Prior to this change Travis runs are broken because of
changes to the Ruby ecosystem. See the ticket FM-7655 for details.


metadata.json
```diff
@@ -1,45 +1,45 @@
...
+    }
+  ],
+  "operatingsystem_support": [
+    {
+      "operatingsystem": "windows",
+      "operatingsystemrelease": [
+        "Server 2008 R2",
+        "Server 2012",
+        "Server 2012 R2",
+        "Server 2016",
```
ThoughtCrhyme: no 2019?

RandomNoun7: Seems like we should ticket that, because yeah it should be there.

PR #104, opened 2019-04-11
(MODULES-8856) Redact debug

michaeltlombardi:
Prior to this commit the provider would output the full Powershell log to be executed during debug runs. This is to ensure that we can verify exactly what will be executed. However, this also meant that credentials and other secrets could leak into the debug output, and thus also into PuppetDB logs.

This commit modifies the provider, ERB templates, and a helper file to ensure that methods for formatting Puppet data to be PowerShell compliant are able to take a new parameter, `redact`.

Passing this parameter will cause the script builder to redact any sensitive data it is passed. This flag is forwarded along through several helpers. This implementation allows us to reuse the code for interpolating the correct values into the PowerShell script for the actual execution and for the debug messaging.

There is possibly a better implementation for this change.

This commit also includes updated spec and integration tests to verify the behavior when choosing to redact sensitive values.

### DIFF
--- a/lib/puppet/provider/base_dsc_lite/powershell.rb
+++ b/lib/puppet/provider/base_dsc_lite/powershell.rb
@@ -77,7 +77,7 @@ EOT
     version = Facter.value(:powershell_version)
     Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
-    Puppet.debug "\n" + script_content
+    Puppet.debug "\n" + self.class.redact_content(script_content)
 
     if !PuppetX::DscLite::PowerShellManager.supported?
       self.class.upgrade_message
@@ -97,7 +97,7 @@ EOT
 
   def create
     script_content = ps_script_content('set')
-    Puppet.debug "\n" + script_content
+    Puppet.debug "\n" + self.class.redact_content(script_content)
 
     if !PuppetX::DscLite::PowerShellManager.supported?
       self.class.upgrade_message
@@ -148,7 +148,7 @@ EOT
     when dsc_value.class.name == 'Hash'
       "@{" + dsc_value.collect{|k, v| format_dsc_value(k) + ' = ' + format_dsc_value(v)}.join('; ') + "}"
     when dsc_value.class.name == 'Puppet::Pops::Types::PSensitiveType::Sensitive'
-      "'#{escape_quotes(dsc_value.unwrap)}'"
+      "'#{escape_quotes(dsc_value.unwrap)}' # PuppetSensitive"
     else
       fail "unsupported type #{dsc_value.class} of value '#{dsc_value}'"
     end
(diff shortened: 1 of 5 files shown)


michaeltlombardi:
These changes have passed in AdHoc.

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -77,7 +77,7 @@ def exists?
     version = Facter.value(:powershell_version)
     Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
-    Puppet.debug "\n" + script_content
+    Puppet.debug "\n" + ps_script_content('test', true)
```
Iristyle: Rather than running the template twice and calling all of the callsites, how hard would it be to just enumerate the Sensitive values we've been given and just `.gsub` across the generated blob of text?

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -144,29 +144,29 @@ def self.format_dsc_value(dsc_value)
...
       "$#{dsc_value.to_s}"
     when dsc_value.class.name == 'Array'
-      "@(" + dsc_value.collect{|m| format_dsc_value(m)}.join(', ') + ")"
+      "@(" + dsc_value.collect{|m| format_dsc_value(m, redact)}.join(', ') + ")"
     when dsc_value.class.name == 'Hash'
-      "@{" + dsc_value.collect{|k, v| format_dsc_value(k) + ' = ' + format_dsc_value(v)}.join('; ') + "}"
+      "@{" + dsc_value.collect{|k, v| format_dsc_value(k, redact) + ' = ' + format_dsc_value(v, redact)}.join('; ') + "}"
     when dsc_value.class.name == 'Puppet::Pops::Types::PSensitiveType::Sensitive'
-      "'#{escape_quotes(dsc_value.unwrap)}'"
+      redact ? "'[REDACTED]'" : "'#{escape_quotes(dsc_value.unwrap)}'"
```
Iristyle: Alternative to my first suggestion... what if we always wrapped such strings in such a way that made them easily removable from the generated template, but that was completely valid PowerShell?

Then you don't really need to keep track of anything at all when doing the replace - you can simply search / replace everything by the pattern.

Iristyle: Given each of these values is on a newline, I think something like this would work:

```powershell
"'#{escape_quotes(dsc_value.unwrap)}' # PuppetSensitive"
```

Then you can just regex find all strings that end like `# PuppetSensitive$` and replace with whatever you want in the debug output, like `"'[REDACTED]'"`. You could make the string more unique by adding a well-known GUID if you were considered about uniqueness (but since it's Debug output I wouldn't be toooooo concerned)

Then a good negative test ensures that you don't ever find the value `PuppetSensitive` in the debug output.

Iristyle:
👍  This seems like a far less intrusive approach that will scale better should we add additional formatting logic, etc

PR #111, opened 2019-06-14
(maint) rubocop cleanup & pdksync

tphoney:


tphoney:
image
running through ad-hoc

spec/acceptance/basic_dsc_resources/failing_dsc_resources_spec.rb
```diff
@@ -26,20 +25,20 @@ module => '#{installed_path}/1.0',
...
-        assert_match(/Stage\[main\]\/Main\/Dsc\[good_resource\]\/ensure\: invoked/, result.stdout, 'DSC Resource missing!')
+        assert_match(%r{Stage\[main\]\/Main\/Dsc\[good_resource\]\/ensure\: invoked}, result.stdout, 'DSC Resource missing!')
       end
     end
   end
 
   context 'FM-2624 - Apply DSC Resource Manifest with Multiple Failing DSC Resources' do
-    throw_message_1 = SecureRandom.uuid
-    throw_message_2 = SecureRandom.uuid
+    throw_message_a = SecureRandom.uuid
```
ThoughtCrhyme: It is not immediately obvious to me why "a" and "b" are preferred over "1" and "2"

spec/lib/dsc_utils.rb
```diff
@@ -43,7 +43,7 @@ def locate_dsc_module(host)
   end
 
   # Return nothing if module is not installed.
-  return ''
+  ''
```
ThoughtCrhyme: I prefer explicit returns, but with the above comment I think this is fine. It just weirds me out to see a bare '' sitting all alone on a line.

spec/acceptance/basic_dsc_resources/failing_dsc_resources_spec.rb
tphoney: numbered variables are a bad code smell, i changed it to letters. I know it's probably the most horrific display of pedantry. But consistency is the name of the game.  https://www.rubydoc.info/gems/rubocop/0.45.0/RuboCop/Cop/Style/VariableNumber

spec/lib/dsc_utils.rb
tphoney: explicit returns are not a 'thing' in ruby. It took me a while to get used to it. What ever is on the last line of a method is returned and something is always returned.

ThoughtCrhyme: Yeah... I know... it still just gives me the heebee-jeebees.

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
```diff
@@ -55,106 +57,101 @@ def self.munge_embeddedinstance(mof_type, embeddedinstance_value)
...
 
           values = values.map { |v| v.to_i }
           unless values.all? { |v| (min <= v) && (v <= max) }
-            fail "#{name} of #{embeddedinstance_name} is outside the valid range of values: #{min} to #{max}: invalid value #{value}"
+            raise "#{name} of #{embeddedinstance_name} is outside the valid range of values: #{min} to #{max}: invalid value #{value}"
           end
 
           if mof_type[:values] && !values.all? { |v| mof_type[:values].include?(v) }
-            fail("Invalid value #{value}. Valid values are #{mof_type[:values].join(', ')}")
+            raise("Invalid value #{value}. Valid values are #{mof_type[:values].join(', ')}")
```
RandomNoun7: This `raise` call doesn't match the above. It has parens around the string. Do we care?

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
```diff
@@ -55,106 +57,101 @@ def self.munge_embeddedinstance(mof_type, embeddedinstance_value)
...
         else
           validation_method = "validate_#{mof_type[:type]}"
-          if respond_to?(validation_method)
-            send(validation_method, name, value)
-          else
-            fail "Unable to validate property #{name} (type #{mof_type[:type]}) of #{embeddedinstance_name}: value #{value}"
-          end
+
+          raise "Unable to validate property #{name} (type #{mof_type[:type]}) of #{embeddedinstance_name}: value #{value}" unless respond_to?(validation_method)
+          send(validation_method, name, value)
```
RandomNoun7: I'm trying to understand why this line is equivalent to the previous if/else statement. Is the intention here that if the `raise` statement is fired that execution should stop and the `send` would not be hit? If execution is allowed to continue after the `raise` fires and `send` is not skipped, then this rearragement would not be equivalent to the prior way the code was arranged.

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
```diff
@@ -170,7 +167,7 @@ def self.ensure_reboot_relationship(resource, pending_relationships)
         if should_add_reboot_relationship(resource, reboot_resource, pending_relationships)
           # otherwise build resource[name] => Reboot[dsc_reboot]
           edge = Puppet::Relationship.new(resource, reboot_resource,
-              { :callback => :refresh, :event => :ALL_EVENTS })
+                                          callback: :refresh, event: :ALL_EVENTS)
```
RandomNoun7: I know that Rubocop is trying to get us to use this new syntax you can use when the final parameter to a function is a hash, but this is objectively worse. `callback:` does not look like a symbol anymore, and it's difficult to tell that this new way of expressing `callback:` and `event:` ends up a single has parameter. It looks, based on this change, like the signature of this method has changed from three parameter to four. I don't know how much support I'll get for this, which is why I'm leaving this as a comment and not a requested change, but can we consider turning this rule off somehow?

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -87,10 +86,10 @@ def exists?
     end
     Puppet.debug "Dsc Resource returned: #{output}"
     data = JSON.parse(output)
-    fail(data['errormessage']) if !data['errormessage'].empty?
+    raise(data['errormessage']) unless data['errormessage'].empty?
     exists = data['indesiredstate']
     Puppet.debug "Dsc Resource Exists?: #{exists}"
-    Puppet.debug "dsc_ensure: #{resource[:dsc_ensure]}" if resource.parameters.has_key?(:dsc_ensure)
+    Puppet.debug "dsc_ensure: #{resource[:dsc_ensure]}" if resource.parameters.key?(:dsc_ensure)
```
RandomNoun7: Even the author of Rubocop agree's that this is a bad rule and specifically patched Rubocop to allow preferring the `has_` syntax. https://github.com/pixta-dev/pixta-rubocop/pull/11

.gitignore
```diff
@@ -1,24 +1,26 @@
...
-/log/
-/pkg/
-/spec/fixtures/manifests/
-/spec/fixtures/modules/
-/tmp/
-/vendor/
-/convert_report.txt
-/update_report.txt
-.DS_Store
+.git/
```
RandomNoun7: The line endings on this file have changed. I don't disagree that LF line endings might be more appropriate for a .gitignore file, but perhaps we could do it in a maint commit so it doesn't muddy the waters on what's been changed as part of the rubocop and pdksync process?

.travis.yml
```diff
@@ -1,43 +1,44 @@
...
-deploy:
-  provider: puppetforge
-  user: puppet
-  password:
-    secure: ""
-  on:
-    tags: true
-    all_branches: true
-    condition: "$DEPLOY_TO_FORGE = yes"
+---
```
RandomNoun7: Line endings again obscuring the real changes.

Gemfile
```diff
@@ -1,78 +1,78 @@
...
-  File.join(Dir.home, '.gemfile'),
-]
-
-extra_gemfiles.each do |gemfile|
-  if File.file?(gemfile) && File.readable?(gemfile)
-    eval(File.read(gemfile), binding)
-  end
-end
-# vim: syntax=ruby
+source ENV['GEM_SOURCE'] || 'https://rubygems.org'
```
RandomNoun7: Line endings.

metadata.json
```diff
@@ -1,46 +1,46 @@
...
-    "powershell",
-    "desired-state-configuration",
-    "dsc"
-  ],
-  "pdk-version": "1.8.0",
-  "template-url": "https://github.com/puppetlabs/pdk-templates",
-  "template-ref": "heads/master-0-g7281db5"
-}
\ No newline at end of file
+{
```
RandomNoun7: Line endings.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -310,15 +311,15 @@ def close_stream(stream, style = :inprocess)
       expect(result[:exitcode]).to eq(0)
     end
 
-    it "should properly interleave output written natively to stdout and via Write-XXX cmdlets" do
+    it 'properlies interleave output written natively to stdout and via Write-XXX cmdlets' do
```
RandomNoun7: Typo on `properlies`

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -476,209 +476,198 @@ def output_cmdlet
...
+      work_dir = ENV['WINDIR']
       current_work_dir = Dir.getwd
 
-      first_cwd = manager.execute('(Get-Location).Path',nil,work_dir)[:stdout]
+      first_cwd = manager.execute('(Get-Location).Path', nil, work_dir)[:stdout]
       second_cwd = manager.execute('(Get-Location).Path')[:stdout]
 
       # Paths should be case insensitive
       expect(first_cwd.downcase).to eq("#{work_dir}\r\n".downcase)
       expect(second_cwd.downcase).to eq("#{current_work_dir}\r\n".downcase)
```
RandomNoun7: Do we need to add a third test here to ensure that `first_cwd` and `second_cwd` are not equal? This test doesn't quite say what it claims to say if the two happen to the same does it?

spec/integration/puppet_x/puppetlabs/powershell_version_spec.rb
```diff
@@ -1,79 +1,77 @@
-#! /usr/bin/env ruby
+# rubocop:disable RSpec/FilePath
+# rubocop:disable RSpec/AnyInstance
+# ! /usr/bin/env ruby
```
RandomNoun7: In `powershell_hash_formatter_spec.rb` the shebang was removed, which made sense since I don't there's any expectation that we should be able to just invoke this file with ruby on it's own is there?

spec/lib/dsc_utils.rb
```diff
@@ -86,18 +86,17 @@ def setup_dsc_resource_fixture(host)
   scp_to(host, fake_reboot_type_source_path, dsc_type_path)
 
   # if installing to a master, then agents must pluginsync
-  if (host['roles'].include?('master'))
-    step 'Sync DSC resource implementations from master to agents'
-    on(agents, puppet('agent -t --environment production'), :acceptable_exit_codes => [0,2])
-  end
+  return unless host['roles'].include?('master')
+  step 'Sync DSC resource implementations from master to agents'
+  on(agents, puppet('agent -t --environment production'), acceptable_exit_codes: [0, 2])
```
RandomNoun7: The intent of the code seems much clearer under the former arrangement. It took me a little while of staring at this before I was able to work  out how these two sets of code were equivalent. If we revert this to the if block style will Rubocop complain?

spec/unit/puppet/type/base_dsc_spec.rb
```diff
@@ -1,30 +1,25 @@
...
+    expect(base_dsc_lite.to_s).to eq('Base_dsc_lite[foo]')
   end
 
   # Configuration PROVIDER TESTS
 
-  describe "powershell provider tests" do
-
-    it "should successfully instantiate the provider" do
+  describe 'powershell provider tests' do
+    it 'successfullies instantiate the provider' do
```
RandomNoun7: Type on `successfullies`.

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
tphoney: i dont see the difference

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
tphoney: if you feel strongly about this, it is worth making changes in pdk-templates

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
tphoney: this is definitely a separate pr / ticket.

.gitignore
tphoney: this happened because of the pdksync, not because i felt like making the changes.

.travis.yml
tphoney: this happened because of pdksync

Gemfile
tphoney: pdksync againg

lib/puppet/provider/base_dsc_lite/powershell.rb
tphoney: if you want to change the pdk_template. please put up a pr, were we can discuss this in the correct environment

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
tphoney: a raise is an exception, and should be treated as such. the send line, is the last statement in the function, and the actual implicit return

metadata.json
tphoney: came from pdksync

spec/lib/dsc_utils.rb
tphoney: i would argue the opposite.  what happens when the if statement fails, there is no clear return.

tphoney:
re run with the changes 
image

lib/puppet/provider/base_dsc_lite/powershell.rb
RandomNoun7: Can do.

spec/lib/dsc_utils.rb
RandomNoun7: Ruby's implicit returns make it difficult to tell based on either form of this code block where the return should come from. Even on this form, should the return come from line 91? There's another `on` call on line 83, is that one returning data also? Really, neither of them are returning anything, and the intent of this function is not to return at all, but to perform the `step`s like the one on 90.

And the `return unless` doesn't make it clearer what the return should be since it's not really returning anything but being used to halt execution. It just makes it less clear that the two lines below should always be treated as a single unit with the conditional above.

.gitignore
RandomNoun7: Yup, it's one of the issues we have with PDK. It never uses the current system line endings, or inspects the file to decide what it's current line endings are to use that. What we usually do is run the pdk update manually, and then go back and rectify any files like this one that have had the line endings changed, and switch them back prior to merging the commit so that the actual content changes are clearer.

I say this just to make it clear that I'm not accusing you of making capricious or random changes, and if you want to switch the line endings and take the hit of a big diff like that in here, that's a valid choice, it's just not one we've made in the past.

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
RandomNoun7: It'd pedantry, and I notice now too that the difference in call is present in the old version anyway and not worth discussing.

spec/lib/dsc_utils.rb
RandomNoun7: As requested, I would like a Rubocop disable here if possible.

lib/puppet_x/puppetlabs/dsc_lite/dsc_type_helpers.rb
RandomNoun7: A Rubocop disable here as well would be appreciated.

tphoney:
image
Suggested changes made @RandomNoun7

PR #126, opened 2019-08-29
(MODULES-9800) Fix Lib Environment Variable Check

RandomNoun7:
Prior to this commit, the function that checks for valid paths in the
`lib` environment variable was broken. It would return true, and raise
an error if any value at all was assigned to that variable.

This commit fixes the behavior of that function and unit tests it to
prevent regressions.


RandomNoun7:
image

PR #134, opened 2019-11-29
(FM 8425) - Replace Library Code

david22swan:


.sync.yml
```diff
@@ -20,6 +20,9 @@ Gemfile:
     - gem: beaker-testmode_switcher
       version: "~> 0.4"
     - gem: master_manipulator
+  optional:
+    ':development':
+      - gem: 'ruby-pwsh'
```
DavidS: no project should need the gem AND the module dependency. As a module I'd prefer the module dependency (and fixture). If that leads to issues, we need to fix them further upstream.

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -58,14 +57,13 @@ def self.template_path
 
   def self.powershell_args
     ps_args = ['-NoProfile', '-NonInteractive', '-NoLogo', '-ExecutionPolicy', 'Bypass']
-    ps_args << '-Command' unless PuppetX::DscLite::PowerShellManager.supported?
+    ps_args << '-Command' unless Pwsh::Manager.pwsh_supported?
     ps_args
   end
```
DavidS: Below the usage of this method is replaced by `Pwsh::Manager.powershell_args` - if there are no uses left, we can delete it here :interrobang:

.sync.yml
michaeltlombardi: I _think_ this is only for the development purposes, not for the shipped project, but I'm happy to double check whether this is necessary!

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -1,8 +1,7 @@
 require 'pathname'
 require 'json'
 if Puppet::Util::Platform.windows?
-  require_relative '../../../puppet_x/puppetlabs/dsc_lite/powershell_manager'
-  require_relative '../../../puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter'
+  require 'ruby-pwsh'
```
michaeltlombardi: I think, technically, we can include this without the guard, as loading the `ruby-pwsh` library won't blow up on non-Windows.

lib/puppet/provider/base_dsc_lite/powershell.rb
michaeltlombardi: @DavidS correct, this can be dropped in favor of using the helper from the gem.

If we were keeping it though, it should be `Pwsh::Manager.windows_powershell_supported?` I think - `pwsh` is for PowerShell Core, `windows_powershell` prefix for Windows PowerShell, which is required for DSC at this time.

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -76,7 +74,7 @@ def exists?
     script_content = ps_script_content('test')
     Puppet.debug "\n" + self.class.redact_content(script_content)
 
-    if !PuppetX::DscLite::PowerShellManager.supported?
+    if !Pwsh::Manager.pwsh_supported?
```
michaeltlombardi: Should be `windows_powershell_supported?`

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -96,7 +94,7 @@ def create
     script_content = ps_script_content('set')
     Puppet.debug "\n" + self.class.redact_content(script_content)
 
-    if !PuppetX::DscLite::PowerShellManager.supported?
+    if !Pwsh::Manager.pwsh_supported?
```
michaeltlombardi: Should be `windows_powershell_supported?`

Should this be an `unless` statement?

lib/puppet/provider/base_dsc_lite/powershell.rb
```diff
@@ -151,7 +149,7 @@ def self.format_dsc_value(dsc_value)
   end
 
   def self.format_dsc_lite(dsc_value)
-    PuppetX::PuppetLabs::DscLite::PowerShellHashFormatter.format(dsc_value)
+    Pwsh::Util.format_powershell_value(dsc_value)
```
michaeltlombardi: 🤔 this one we may have to do a bit of heavy lifting on as the DSC Lite module had to have additional formatting management for sensitive values and CIM objects.

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
```diff
@@ -1,85 +0,0 @@
...
-            format_number(dsc_value)
-          elsif [:true, :false].include?(dsc_value)
-            format_boolean(dsc_value)
-          elsif ['trueclass', 'falseclass'].include?(dsc_value.class.name.downcase)
-            "$#{dsc_value}"
-          elsif dsc_value.class.name == 'Array'
-            format_array(dsc_value)
-          elsif dsc_value.class.name == 'Hash'
-            format_hash(dsc_value)
-          elsif dsc_value.class.name == 'Puppet::Pops::Types::PSensitiveType::Sensitive'
```
michaeltlombardi: We may need to monkey patch this method to retain handling sensitive objects.

lib/puppet_x/puppetlabs/dsc_lite/powershell_hash_formatter.rb
```diff
@@ -1,85 +0,0 @@
...
-
-        private_class_method def self.format_boolean(value)
-          "$#{value}"
-        end
-
-        private_class_method def self.format_array(value)
-          '@(' + value.map { |m| format(m) }.join(', ') + ')'
-        end
-
-        private_class_method def self.format_hash(value)
```
michaeltlombardi: Similarly, we may need to borrow from this  and `format_cim_instance` for hashes which are special DSC object types, like credentials or other CIM instances.

.sync.yml
michaeltlombardi: Tested this tonight and the fixtures will help with `pdk bundle exec puppet ...` calls but not `pdk test unit` - it will blow up on the requires statement in the provider.

DavidS: that is weiiiird. seems like something in the rspec-puppet chain is not setting up load paths correctly.

lib/puppet/provider/base_dsc_lite/powershell.rb
david22swan: Removed and replaced

lib/puppet/provider/base_dsc_lite/powershell.rb
david22swan: Done

lib/puppet/provider/base_dsc_lite/powershell.rb
david22swan: Updated

lib/puppet/provider/base_dsc_lite/powershell.rb
david22swan: updated, also made it a strait if and swapped the content

david22swan: Just read better that way to me

lib/puppet/provider/base_dsc_lite/powershell.rb
david22swan: What do you mean?

lib/puppet/provider/base_dsc_lite/powershell.rb
michaeltlombardi: Unlike practically all other powershell-dependent modules, invoking DSC requires that we handle some very special objects with particular types not matched in ruby, namely CIM instances.

We also need to write a specific handler for converting from a sensitive Puppet object to the appropriate PowerShell object representation.

lib/puppet/provider/base_dsc_lite/powershell.rb
michaeltlombardi: Unlike practically all other powershell-dependent modules, invoking DSC requires that we handle some very special objects with particular types not matched in ruby, namely CIM instances.

We also need to write a specific handler for converting from a sensitive Puppet object to the appropriate PowerShell object representation.

david22swan: Ok, have returned `powershell_hash_formatter.rb`, cleaned out everything in it that's mirrored in the gem and added in a call to said gem.
So it will check if it's a `Hash` or a `Puppet::Pops::Types::PSensitiveType::Sensitive` and apply the appropriate logic if it is. And if it isn't it will pass it on though to the gems format method.

michaeltlombardi:
Is it just me or is the appveyor run not executing acceptance tests?

https://ci.appveyor.com/project/puppetlabs/puppetlabs-dsc-lite/builds/29365114

😶

david22swan:
@michaeltlombardi PDKSync had stripped the acceptance testing out. Have added back in

PR #188, opened 2022-01-24
(docs) Add messaging about the differences between this and the dsc modules

binford2k:


README.md
```diff
@@ -24,9 +24,30 @@
 
 ## Module description
 
-The Puppet dsc_lite module allows you to manage target nodes using Windows PowerShell DSC (Desired State Configuration) Resources.
+The Puppet `dsc_lite` module allows you to manage target nodes using arbitrary Windows PowerShell DSC (Desired State Configuration) Resources.
 
-**Note that using dsc_lite requires previous experience with DSC and PowerShell, unlike our other modules that are easier to get started with. It is an alternative approach to Puppet's dsc module, providing more flexibility for advanced users.**  
+### Warning:
+
+***Using dsc_lite requires advanced experience with DSC and PowerShell, unlike our other modules that are far easier to use. It is an alternative approach to Puppet's family of [DSC modules](http://forge.puppet.com/dsc), providing more flexibility for certain niche use cases. There are many drawbacks to this approach and we highly recommend that you use the existing family of [DSC modules](http://forge.puppet.com/dsc) instead.***
```
michaeltlombardi, line 31:
```suggestion
***Using dsc_lite requires advanced experience with DSC and PowerShell, unlike our other modules that are far easier to use. It is an alternative approach to Puppet's family of [DSC modules](http://forge.puppet.com/dsc), providing more flexibility for certain niche use cases. There are many drawbacks to this approach and we highly recommend that you use the existing family of [DSC modules](http://forge.puppet.com/dsc) instead, where possible.***
```

README.md
```diff
@@ -24,9 +24,30 @@
...
+
+The `dsc_lite` module contains a lightweight `dsc` type, which a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. This type does not contain any DSC resources itself, but can invoke arbitrary DSC Resources that already exist on the managed node. Much like the `exec` type, it simply passes parameters through to the underlying DSC Resource without any validation.
```
michaeltlombardi, line 33:
```suggestion
The `dsc_lite` module contains a lightweight `dsc` type, which is a streamlined and minimal representation of a DSC Resource declaration in Puppet syntax. This type does not contain any DSC resources itself, but can invoke arbitrary DSC Resources that already exist on the managed node. Much like the `exec` type, it simply passes parameters through to the underlying DSC Resource without any validation.
```

README.md
```diff
@@ -24,9 +24,30 @@
...
+
+You should *only* use the `dsc_lite` module if either of these cases apply to you:
+
+* You need to use multiple versions of the same DSC resource
+* You need to use a DSC resource that isn't published to the [Puppet Forge](http://forge.puppet.com/dsc).
+    * If you find a DSC Resource that hasn't been automatically imported, it's very likely due to the original DSC Resource failing schema validation. The `dsc_lite` module can get you by, but you should report the error upstream so that the original author can correct their code.
+    * If you have custom DSC Resources, you can use the [Puppet.dsc module builder](https://github.com/puppetlabs/Puppet.Dsc) to build your own Puppet module from it.
```
michaeltlombardi, line 48:
```suggestion
You should *only* use the `dsc_lite` module when any of these cases apply to you:

* You need to use multiple versions of the same DSC resource
* The upstream DSC Resource's implementation does not match its declared API but is usable otherwise and needed.
  * In this case, `dsc_lite` should be treated as a stop-gap solution until the upstream DSC Resource can be patched to fix the misimplementation.
* You need to use a DSC Resource that isn't published to the [Puppet Forge](http://forge.puppet.com/dsc).
    * If you find a DSC Resource that hasn't been automatically imported, it's very likely due to the original DSC Resource failing schema validation. The `dsc_lite` module can get you by, but you should report the error upstream so that the original author can correct their code.
    * If you have custom DSC Resources, you can use the [Puppet.dsc module builder](https://github.com/puppetlabs/Puppet.Dsc) to build your own Puppet module from it.
```

michaeltlombardi (changes requested):
Some minor suggestions but this is A++ helping us move to more consistent/clear language.

README.md
binford2k: How would a user know about the API mismatch? I struggled with this a bit before settling on the "not on the forge" phrasing.

I suppose that if someone had the expertise to use dsc_lite, they'd know what that meant?

michaeltlombardi: Yeah, functionally, if you don't know what this means, you don't need to use `dsc_lite` to resolve it. This is a case where like the type does not correctly report itself even internally (ie the `Test` method is unreliable) or the API is mismatched.

README.md
```diff
@@ -24,9 +24,32 @@
...
+
+You should *only* use the `dsc_lite` module when any of these cases apply to you:
+
+* You need to use multiple versions of the same DSC resource
+* The upstream DSC Resource's implementation does not match its declared API but is usable otherwise and needed.
+  * In this case, `dsc_lite` should be treated as a stop-gap solution until the upstream DSC Resource can be patched to fix the misimplementation.
+* You need to use a DSC Resource that isn't published to the [Puppet Forge](http://forge.puppet.com/dsc).
+    * If you find a DSC Resource that hasn't been automatically imported, it's very likely due to the original DSC Resource failing schema validation. The `dsc_lite` module can get you by, but you should report the error upstream so that the original author can correct their code.
```
binford2k, line 49:
```suggestion
    * If you find a DSC Resource that hasn't been automatically imported, it's very likely due to an API mismatch as described above.
```

binford2k: @michaeltlombardi is this accurate?

michaeltlombardi: yup!

michaeltlombardi (approved):

PR #201, opened 2022-06-17
Add a simple spec test to validate catalog compilation

binford2k:
This just verifies that deferred functions are compiled into the catalog
for the agent to deal with instead of being rendered into powershell
during compilation on the server.

### DIFF
--- a/spec/unit/puppet/type/dsc_spec.rb
+++ b/spec/unit/puppet/type/dsc_spec.rb
@@ -4,6 +4,14 @@ require 'spec_helper'
 require 'puppet/type'
 require 'puppet/type/dsc'
 
+# borrowed from
+# https://github.com/puppetlabs/puppet/blob/363f22b2a4125bedb88921d64b277d6f8adbbef2/spec/lib/puppet_spec/compiler.rb#L4-L8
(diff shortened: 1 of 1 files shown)


binford2k:
@tvpartytonight @Iristyle I think this is validating what I think it is. Would you agree?

spec/unit/puppet/type/dsc_spec.rb
```diff
@@ -171,5 +179,18 @@
...
+        dsc {'foo':
+          resource_name => 'WindowsFoo',
+          properties    => {
+            password => Deferred("join", [[1,2,3], ':']),
+          }
+        }
+      END
+
+      expect(catalog.resource(:dsc, 'foo')[:properties]['password'].arguments).to eq([[1, 2, 3], ':'])
+    end
```
joshcooper, line 194: I think this is correct (the catalog contains an instance of a deferred object with certain arguments). One nit, the test description says it "resolves deferred values"  but really that doesn't occur until `Puppet::Resource::Catalog#to_ral` is called (by dfeault) or when the resource is evaluated (when `Puppet[:preprocess_deferred] = false`)

joshcooper (approved):

PR #207, opened 2024-05-09
(CAT-1776) Implement reusable workflows

LukasAud:
This PR aims to bring the DSC_lite tool workflow structure in line with the rest of the DevX toolset. This change implements some of the modules reusable workflows for consistency, as we officially recognise DSC_lite as having a module-like structure and function. Here is a list of changes made:

- Removed the `auto_release.yml` workflow as it is no longer needed
- Implement the reusable release_prep workflow to replace it
- Removed the `spec.yml` workflow as it is no longer needed
- Removed the `stale.yml` workflow as it is no longer needed
- Implement the reusable mend workflow to collect internal data
- Adjust the reusable labeller workflow to label PRs


.github/workflows/mend.yml
```diff
@@ -0,0 +1,16 @@
...
+      - synchronize
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+jobs:
+
+  mend:
+    uses: "puppetlabs/cat-github-actions/.github/workflows/tooling_mend_ruby.yml@main"
+    secrets: "inherit"
```
jordanbreen28, line 16: missing newline

.sync.yml
```diff
@@ -15,6 +15,8 @@ spec/spec_helper.rb:
   unmanaged: false
 .github/workflows/ci.yml:
   unmanaged: false
+.github/workflows/release_prep.yml:
```
jordanbreen28, line 18: should we remove the rest from here now?
https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/45c89bdb221e96d62695e3490f5cc539c9074942/.sync.yml#L24C1-L30C19


Should probably add the mend one here too what do you think?

jordanbreen28 (approved):

PR #218, opened 2024-07-16
(CAT-1927) - Remove Windows 7, 8.x and 2008 R2 Support

jordanbreen28:
Remove Windows 7, 8.x and 2008 R2 Support, these platforms are no longer supported by puppet agent and therefore cannot be supported for use with this module.


LukasAud (approved):
LGTM

PR #224, opened 2025-01-15
(MAINT) Addressing undetected legacy facts

LukasAud:
A recent investigation revealed the presence of untreated legacy facts in the module. This commit addresses the issue by updating them to their modern counterparts.


gavindidrichsen (approved):
Good catch @LukasAud !

PR #230, opened 2025-04-09
(CAT-2281) Remove puppet 7 infrastructure

LukasAud:
Puppet 7 is EOL. Therefore, we can remove the test infrastructure for it. This PR aims to clear up any testing/config infrastructure related to Puppet 7 and, by extension, Ruby 2.7.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 /spec/fixtures/modules/*
 /tmp/
 /vendor/
+/.vendor/
 /convert_report.txt
 /update_report.txt
 .DS_Store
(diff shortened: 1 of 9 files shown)


gavindidrichsen (changes requested):
Hi @LukasAud .  If you're removing the workflow restarter on this, then you can probably remove https://github.com/puppetlabs/puppetlabs-dsc_lite/blob/CAT-2281-Remove_puppet_7_test_infrastructure/.github/workflows/workflow-restarter.yml too

jordanbreen28 (approved):

gavindidrichsen (approved):

PR #233, opened 2025-09-30
(CAT-2453) Update rubocop to remove rexml dependency

david22swan:
The version of rexml brought in by the current rubocop versions has a security vulnerability.
Updating to these rubocop versions will remove their dependency entirely.
Adding in plugins that where previously part of the default package, but have been split off since our previously used version.


pmcmaw (approved):

pmcmaw:
TWINGATE failure, merging.

